PR #2, opened 2026-02-10
fix: repair updateWebhook and deleteWebhook MCP tools

nathanliow:
**Summary**       
  - updateWebhook was sending a partial object to the Helius API, which requires the full webhook object on PUT. It also lacked 
  webhookType in its schema entirely, so it could never pass that required field. Fixed by auto-fetching the existing webhook
  and merging caller-provided fields on top, and adding webhookType to the Zod schema.                                          
  - deleteWebhook was failing with "Unexpected token 'n', "null" is not valid JSON" because restRequest always sets
  Content-Type: application/json, but DELETE sent no body — causing the Helius API to try parsing null as JSON. Fixed by sending
   an empty JSON body {}.

  **Root causes**
  **updateWebhook (HTTP 400)**
  1. webhookType was missing from the Zod schema, so it was never accepted as a parameter and never sent to the API
  2. The handler only included fields the caller explicitly provided, sending a partial object — but Helius PUT /v0/webhooks/:id
   requires the full webhook object (webhookURL, webhookType, accountAddresses are all mandatory)

  **deleteWebhook (HTTP 400)**
  1. restRequest unconditionally sets Content-Type: application/json on every request
  2. DELETE was sent with no body, so the Helius API received a request with a JSON content type but a null body and failed to
  parse it

  **Test plan**
  - updateWebhook with only webhookID + one changed field (e.g. transactionTypes) succeeds — unchanged fields are preserved from
   existing webhook
  - updateWebhook with explicit webhookType override works
  - deleteWebhook successfully removes a webhook without errors

### DIFF
--- a/helius-mcp/src/tools/webhooks.ts
+++ b/helius-mcp/src/tools/webhooks.ts
@@ -91,16 +91,16 @@ export function registerWebhookTools(server: McpServer) {
       webhookURL: z.string().describe('Your webhook URL endpoint'),
       webhookType: z.enum(['enhanced', 'raw', 'discord']).describe('Webhook type - use "enhanced" for parsed transaction data with descriptions'),
       accountAddresses: z.array(z.string()).describe('Array of Solana addresses to monitor (up to 100,000 per webhook)'),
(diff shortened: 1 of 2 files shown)


fomichfm (approved):
lgtm

PR #6, opened 2026-02-12
Add docs integration and cleanup

hetdagli234:
- Add docs lookup tools (lookupHeliusDocs, listHeliusDocTopics, getHeliusCreditsInfo)
- Add plan/guide tools
- Fix URLs and credit costs
- Cleanup: remove emoticons, use mainnet
- Bump to v0.5.2


helius-mcp/src/tools/plans.ts
nathanliow, line 1: shouldn't we point to the llms.txt for plans as well? and if we don't have one, we should probably add because it would be much easier to maintain rather than adding an additional area that needs to be updated if the Helius plans change

helius-mcp/src/tools/docs.ts
```diff
@@ -0,0 +1,243 @@
...
+                tableLines.push(lines[j]);
+              }
+              j++;
+              if (j - i > 20) break; // Safety limit
+            }
+            break;
+          }
+        }
+
+        const result = [
```
nathanliow, line 213: intended to be fallback or hardcoded?

helius-mcp/src/tools/plans.ts
hetdagli234: Right, we don't have it, will ask @maxim-schmidt to add it.

nathanliow (approved):
LGTM

If possible, RATE_LIMITS and CREDIT_COSTS should probably be fetched too

nathanliow (approved):

PR #13, opened 2026-02-26
feat(cli): Explicitly Set Project ID on CLI Creation

0xIchigo:
This PR saves the project ID to `~/.helius/config.json` during signup, alongside the JWT and API key. It was previously omitted, so using `helius config` showed `Project ID: not set` even after a successful signup.


fomichfm (approved):
lgtm

PR #15, opened 2026-02-26
fix: pass customer info through upgradePlan and CLI upgrade

nathanliow:
## Summary
- **MCP** (`helius-mcp/src/tools/auth.ts`): remove dead `executeUpgrade` import, add all-or-none customer info validation in `upgradePlan` handler
- **CLI** (`helius-cli/src/lib/checkout.ts`): remove `executeUpgrade` wrapper and import, hardcode `CLI_USER_AGENT` in `executeCheckout` wrapper (no longer accepts a `userAgent` param)
- **CLI** (`helius-cli/src/commands/upgrade.ts`): replace `executeCheckout`/`executeUpgrade` branching with single `executeCheckout` call, add all-or-none customer info validation and basic email format validation

All three surfaces (SDK, MCP, CLI) now consistently use `executeCheckout` for upgrades with proper customer info validation.

Depends on helius-labs/helius-sdk#294

## Test plan
- [x] `pnpm build` in `helius-mcp/` — builds clean
- [x] `pnpm build` in `helius-cli/` — builds clean
- [ ] Test partial customer info (e.g. `--first-name` alone) — errors early
- [ ] Test bad email format — errors early
- [ ] Test full customer info or no customer info — proceeds to checkout


helius-cli/src/commands/upgrade.ts
```diff
@@ -135,14 +138,31 @@ export async function upgradeCommand(options: UpgradeOptions): Promise<void> {
...
+          {
+            plan: planKey,
+            period: options.period,
+            refId: project.id,
+            couponCode: options.coupon,
+            email: options.email,
+            firstName: options.firstName,
+            lastName: options.lastName,
+          },
+          undefined,
```
0xIchigo: The CLI should always default to `CLI_USER_AGENT`, making the param unnecessary

0xIchigo (changes requested):
Very based! Just have a few comments that should be addressed:
- The MCP always calls `executeCheckout`. The CLI branches between `executeCheckout` and `executeUpgrade` based on whether any customer info was supplied. We need to align both the CLI and MCP
- `hasCustomerInfo` triggers on any one of the three fields, so `--first-name John` alone (no email, no last name) routes to `executeCheckout`. We need to double-check about partial customer info validation
- Email validation only happens in the MCP while the CLI accepts any string passed to `--email`

0xIchigo (approved):
Based 🔥

PR #19, opened 2026-02-27
feat(mcp): Make All Data-bearing Guide Tools Fetch Directly From the Docs

0xIchigo:
The MCP currently has two sources of truth—live docs fetched from GitHub, and a parallel copy of hardcoded data. This PR makes it so we now fetch everything directly from the docs/ We also fetch from the new agents section, which covers CLI signup, MCP integrations, and SDK patterns


nathanliow (approved):
LGTM, claude says these tools are duplicative but I think it might be more helpful/clear to the LLM imo

PR #29, opened 2026-02-28
fix(cli): Improve Signup UX

0xIchigo:
This PR implements several quick wins to improve the signup UX. Namely,
- Fix SOL amount inconsistency (~0.01 -> ~0.001) in keygen
- `helius signup` now auto-generates a keypair if none exists instead of erroring
- `helius signup` checks SOL/USDC balance before attempting payment, shows exactly what's missing
- `helius signup` now shows full API key + RPC endpoints for existing accounts (previously just said "saved")
- `helius config show --reveal` shows the full API key instead of truncating

### DIFF
--- a/helius-cli/src/commands/signup.ts
+++ b/helius-cli/src/commands/signup.ts
@@ -102,7 +102,13 @@ export async function signupCommand(options: SignupOptions): Promise<void> {
         }
       }
       if (result.apiKey) {
-        console.log(chalk.green(`\nAPI key saved to ${SHARED_CONFIG_PATH}`));
+        console.log(`\nAPI Key: ${chalk.cyan(result.apiKey)}`);
+        console.log(chalk.green(`Saved to ${SHARED_CONFIG_PATH}`));
+      }
+      if (result.endpoints) {
+        console.log(chalk.bold("\nRPC Endpoints:"));
+        console.log(`  Mainnet: ${chalk.blue(result.endpoints.mainnet)}`);
+        console.log(`  Devnet:  ${chalk.blue(result.endpoints.devnet)}`);
       }
       console.log(chalk.gray("\nNo payment required. Use `helius projects` to view details."));
       return;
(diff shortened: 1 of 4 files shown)


helius-cli/src/commands/signup.ts
```diff
@@ -32,20 +33,49 @@ export async function signupCommand(options: SignupOptions): Promise<void> {
...
+    spinner?.succeed(`Wallet loaded: ${walletAddress}`);
+
+    // Check balance before attempting payment
+    spinner?.start("Checking wallet balance...");
+    const solBalance = await checkSolBalance(walletAddress);
+    const usdcBalance = await checkUsdcBalance(walletAddress);
+    const solAmount = Number(solBalance) / 1_000_000_000;
+    const usdcAmount = Number(usdcBalance) / 1_000_000;
+    const solOk = solBalance >= 1_000_000n;    // ~0.001 SOL
+    const usdcOk = usdcBalance >= 1_000_000n;  // 1 USDC minimum
```
nathanliow: shouldnt be hardcoding 1 USDC, users can signup for dev/business/pro plan with helius signup so preflight check doesn't make sense when a user signs up for dev plan with 2 USDC. maybe can check against PLAN_CATALOG

helius-cli/src/commands/signup.ts
```diff
@@ -1,11 +1,12 @@
...
 import ora from "ora";
-import { loadKeypairFromFile } from "../lib/wallet.js";
+import { loadKeypairFromFile, getAddress } from "../lib/wallet.js";
 import { agenticSignup, listProjects } from "../lib/api.js";
 import { setJwt, setApiKey, setSharedApiKey, setProjectId, SHARED_CONFIG_PATH } from "../lib/config.js";
-import { keypairExists } from "./keygen.js";
+import { keypairExists, keygenCommand } from "./keygen.js";
 import { formatEnumLabel } from "../lib/formatters.js";
 import { outputJson, exitWithError, ExitCode, type OutputOptions } from "../lib/output.js";
+import { checkSolBalance, checkUsdcBalance } from "helius-sdk/auth/checkBalances";
```
nathanliow: src/lib/payment.ts already re-exports these. Should use ../lib/payment.js to stay consistent with the project's convention of going through local re-exports

nathanliow (changes requested):

nathanliow (approved):
LGTM

PR #30, opened 2026-02-28
feat(mcp + cli): Add Solana Knowledge Tools

0xIchigo:
This PR adds Solana knowledge tools to the MCP so we can fetch SIMDs, read files from Agave & FD's GitHub repos, search the official Solana docs, and fetch some of our Helius blog posts. We also added a new `helius simd` command to the CLI to streamline fetching SIMDs

### DIFF
--- a/helius-mcp/src/tools/solana-knowledge.ts
+++ b/helius-mcp/src/tools/solana-knowledge.ts
@@ -462,43 +462,62 @@ export function registerSolanaKnowledgeTools(server: McpServer) {
         const queryLower = query.toLowerCase();
         const queryWords = queryLower.split(/\s+/).filter((w) => w.length > 2);
         const lines = content.split('\n');
-        const relevantSections: string[] = [];
-        let inRelevantSection = false;
-        let sectionDepth = 0;
-        let sectionLines: string[] = [];
 
-        for (const line of lines) {
-          const headerMatch = line.match(/^(#{1,4})\s+(.+)/);
-
-          if (headerMatch) {
-            const depth = headerMatch[1].length;
-            const title = headerMatch[2].toLowerCase();
-
-            // Flush previous relevant section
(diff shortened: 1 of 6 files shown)


helius-mcp/src/tools/solana-knowledge.ts
```diff
@@ -0,0 +1,661 @@
...
+        };
+      }
+    }
+  );
+
+  // =========================================================================
+  // Tool 4: searchSolanaDocs — Fetch official Solana documentation
+  // =========================================================================
+  server.tool(
+    'searchSolanaDocs',
```
nathanliow, line 445: header-only matching — if the user queries "rent" but the section header says "Accounts" (with rent explained inside the body), it won't match. The fallback (line-level grep) helps but returns decontextualized lines rather than sections

helius-mcp/src/tools/solana-knowledge.ts
```diff
@@ -0,0 +1,661 @@
...
+        const relevantSections: string[] = [];
+        let inRelevantSection = false;
+        let sectionDepth = 0;
+        let sectionLines: string[] = [];
+
+        for (const line of lines) {
+          const headerMatch = line.match(/^(#{1,4})\s+(.+)/);
+
+          if (headerMatch) {
+            const depth = headerMatch[1].length;
```
nathanliow: Don't flush the outer section when a deeper matching header is encountered

One approach: only flush when depth <= sectionDepth AND the new header doesn't match the query — otherwise just continue accumulating.

nathanliow (changes requested):
Think `getSIMD`, `readSolanaSourceFile`, `searchSolanaDocs`, and `fetchHeliusBlog` need to all be added to routing guidance

nathanliow (approved):
LGTM

PR #45, opened 2026-03-04
feat(plugin): Make Helius Cursor Plugin

0xIchigo:
This PR adds a Cursor marketplace plugin that mirrors our Claude Code plugin. It includes auto-starting the relevant MCP servers, four skills, five rule files, and a routing agent. We also extend `sync-check.yml` to validate all the reference files found in the new `helius-cursor/` dir


nathanliow (approved):
LGTM

PR #50, opened 2026-03-09
feat(cli): Add `--wait` Flag to `helius signup`

0xIchigo:
This PR adds a `--wait` flag to `helius signup` that polls wallet balances every 5s when funds are insufficient, auto-continuing signup once SOL + USDC arrive or 5 minutes pass

### DIFF
--- a/helius-cli/bin/helius.ts
+++ b/helius-cli/bin/helius.ts
@@ -98,6 +98,7 @@ program
   .option("--last-name <name>", "Last name (required for paid plans)")
   .option("--discovery-path <text>", "How did you discover Helius?")
   .option("--friction-points <text>", "What friction did you hit finding or setting up Helius?")
+  .option("--wait", "Poll for funds if balance is insufficient, then continue signup automatically")
   .option("--json", "Output in JSON format")
   .action(signupCommand);
 
--- a/helius-cli/src/commands/signup.ts
+++ b/helius-cli/src/commands/signup.ts
@@ -26,15 +26,21 @@ interface SignupOptions extends OutputOptions {
 const POLL_INTERVAL_MS = 5_000;
 const POLL_TIMEOUT_MS = 5 * 60 * 1_000; // 5 minutes
 
+interface FundingResult {
+  funded: boolean;
+  solFunded: boolean;
+  usdcFunded: boolean;
+}
+
 /**
  * Polls SOL and USDC balances until both meet the required thresholds.
- * Returns true when funded, or false if the timeout is reached.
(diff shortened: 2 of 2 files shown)


helius-cli/src/commands/signup.ts
```diff
@@ -20,6 +20,54 @@ interface SignupOptions extends OutputOptions {
...
+async function waitForFunding(
+  walletAddress: string,
+  requiredUsdcRaw: bigint,
+  spinner?: { start(text: string): void; succeed(text: string): void } | null,
+): Promise<boolean> {
+  const start = Date.now();
+  let solFunded = false;
+  let usdcFunded = false;
+
+  while (Date.now() - start < POLL_TIMEOUT_MS) {
```
nathanliow, line 48: we should add try/catch around balance checking loop

helius-cli/src/commands/signup.ts
```diff
@@ -85,16 +133,32 @@ export async function signupCommand(options: SignupOptions): Promise<void> {
...
+        console.error(chalk.gray("\nThen run `helius signup` again, or use `helius signup --wait` to poll until funded."));
+        process.exit(!solOk ? ExitCode.INSUFFICIENT_SOL : ExitCode.INSUFFICIENT_USDC);
+      }
+
+      // Poll until wallet is funded
+      console.log(chalk.gray("\nWaiting for funds... (Ctrl+C to cancel)\n"));
+      const funded = await waitForFunding(walletAddress, requiredUsdcRaw, spinner);
+      if (!funded) {
+        // User cancelled or timeout — should not reach here normally (Ctrl+C exits process)
+        process.exit(!solOk ? ExitCode.INSUFFICIENT_SOL : ExitCode.INSUFFICIENT_USDC);
```
nathanliow: uses the original solOk/usdcOk values from before polling. Have waitForFunding return which assets timed out, or re-check at timeout

helius-cli/bin/helius.ts
```diff
@@ -98,6 +98,7 @@ program
   .option("--last-name <name>", "Last name (required for paid plans)")
   .option("--discovery-path <text>", "How did you discover Helius?")
   .option("--friction-points <text>", "What friction did you hit finding or setting up Helius?")
+  .option("--wait", "Poll for funds if balance is insufficient, then continue signup automatically")
```
nathanliow, line 101: When --json is passed with --wait, the function calls exitWithError on line 136 and never reaches the polling path. Either let it poll (move exitWithError after the !options.wait check), or explicitly reject the combination

nathanliow (approved):
LGTM

PR #51, opened 2026-03-09
feat(cli): Implement Better Error Handling With Custom Catch Block

0xIchigo:
This PR moves the signup command's error handliong from a custom catch block to the shared `handleCommandError` pipeline, so signup errors now get consistent structured classification, `retryable` fields in JSON output, and telemetry like every other command


nathanliow (approved):
LGTM

PR #52, opened 2026-03-09
feat(cli): Improve Messaging For Interrupted Signups

0xIchigo:
This PR detects interrupted signups by checking whether a local API key existed before `agenticSignup` returns `existing_project`. If not, it's a recovery case and shows `"Resuming previous signup - your account was already created."` in green to message this state clearly


nathanliow (approved):
LGTM

PR #54, opened 2026-03-09
fix(cli): Silent Exit in non-JSON Mode

0xIchigo:
This PR aims to fix the issue where `exitWithError()` calls `process.exit()` without printing anything when `json=false` by printing the error message in red with an actionable hint in yellow to match what `handleCommandError()` already does

### DIFF
--- a/helius-cli/src/lib/output.ts
+++ b/helius-cli/src/lib/output.ts
@@ -281,7 +281,7 @@ export function exitWithError(
   errorCode: string,
   message: string,
   details?: Record<string, unknown>,
-  json?: boolean
+  json = true,
 ): never {
   const exitCode = getExitCode(errorCode);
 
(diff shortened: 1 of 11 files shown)


helius-cli/src/lib/output.ts
```diff
@@ -288,6 +288,12 @@ export function exitWithError(
   if (json) {
     const retryable = RETRYABLE_CODES.has(exitCode);
     outputJson({ error: errorCode, message, retryable, ...details });
+  } else {
+    console.error(chalk.red(message));
```
nathanliow, line 292: json is hardcoded to true at every callsite so this never happens, is that intended?

0xIchigo: Damn lmao we should make it so all the callsites pass in `options.json` instead of hardcoding it to `true`, but make it default to `true`. I'll push another commit for this

nathanliow (approved):
LGTM

PR #55, opened 2026-03-10
feat(cli): Add Client-Side Input Validation For Signup, Upgrade, and Webhooks

0xIchigo:
This PR adds new shared validation utils so we now validate input for the `signup`, `upgrade`, and `webhook create/update` commands


helius-cli/src/lib/validation.ts
```diff
@@ -0,0 +1,52 @@
+import { PLAN_CATALOG } from "../lib/checkout.js";
```
nathanliow, line 1: ./checkout.js would work lol

nathanliow (approved):
LGTM, nit is optional

PR #59, opened 2026-03-10
feat(skill): Add Jupiter

0xIchigo:
This PR adds a new skill that combines Jupiter's DeFi APIs with Helius infrastructure. It covers token swaps (Ultra), lending / borrowing (Lend), limit orders (Trigger), DCA (Recurring), perps, prediction markets, token / price data, Token Shield, and the drop-in swap widget (Plugin), all with integration patterns for working with Helius

`helius-skills/helius-jupiter/` is the canonical skill, with support added to our Claude Code plugin, Cursor plugin, and the compiler-generated system prompts for Codex, OpenAI, etc.


0xanmol:
Hey @0xIchigo 

From my earlier audit doc, all issues were addressed, appreciate it. Found three nits during a second pass.

1. Price API v3 response format is incorrect                                                   

`references/jupiter-tokens-price.md` documents the response as:
                                                                                                                                
  ```
const prices = await response.json();
  // Returns: { data: { [mintAddress]: { id, usdPrice, blockId, decimals, priceChange24h, ... } } }                                                                                                 
  const solPrice = prices.data['So11111111111111111111111111111111111111112'];
  ```                                                                                                       

The live API returns mint addresses at the top level — no data wrapper, no id field:
```
  {                                                                                                                                                                                                 
    "So11111111111111111111111111111111111111112": {                                                                                                                                              
      "usdPrice": 82.04,                                                                                                                                                                            
      "blockId": 412733640,
      "decimals": 9,                                                                                                                                                                                
      "priceChange24h": -2.61                                                                                                                                                                     
    }
  }
```                                                        

`prices.data[mint]` returns undefined. Correct access is `prices[mint]`. Since agents generate code from this skill, they'll produce silently broken price lookups.                                   
                                                                                   
2. portal.jup.ag -> developers.jup.ag
   
`portal.jup.ag` 301 redirects to `developers.jup.ag/` (the root homepage, not the portal page). The canonical URL for API key management is now developers.jup.ag/portal. Referenced across `SKILL.md`, `references/jupiter-portal.md`, and `references/jupiter-swap.md`.
                                                                                                                                                                                                    
An agent telling a developer to "get an API key at portal.jup.ag" would redirect them to the correct URL but still is not good for marketing.

3. Quote-only /order response does return requestId

`references/jupiter-swap.md` says:

> // If taker is omitted: no transaction or requestId, just quote data
                                                                                                                          
The live API returns `requestId` even without `taker`, only transaction is `null`. Since this is a skill, the agent will treat this as a rule and tell developers that requestId isn't available in quote-only mode, which is wrong.

nathanliow (approved):
LGTM

PR #60, opened 2026-03-11
fix(mcp): Add Escalating Retry Guidance For Agents

0xIchigo:
This PR makes it so `checkSignupBalance` now tracks consecutive insufficient balance calls and escalates its response:
- Check 1: "Ask the user to send funds. Don't call again until they confirm."
- Check 2: "Still insufficient (2/3). Confirm funds were sent before retrying."
- Check 3+: "Stop polling. Wait for explicit user confirmation."

The counter resets when their balance is sufficient or when `generateKeypair` is called (i.e., fresh flow). This addresses the issue where an agent can loop on `checkSignupBalance` indefinitely because nothing in our responses signals that user intervention is required before retrying

### DIFF
--- a/helius-mcp/src/tools/auth.ts
+++ b/helius-mcp/src/tools/auth.ts
@@ -26,6 +26,10 @@ import { HELIUS_PLANS } from './plans.js';
 
 const PAID_PLAN_ORDER = ['developer', 'business', 'professional'] as const;
 
+/** Tracks consecutive insufficient-balance checks to prevent agent polling loops. */
+let insufficientBalanceChecks = 0;
+const MAX_BALANCE_CHECKS_BEFORE_STOP = 3;
+
 export function registerAuthTools(server: McpServer) {
   // ── Getting Started Guide ──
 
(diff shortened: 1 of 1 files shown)


helius-mcp/src/tools/auth.ts
```diff
@@ -243,22 +250,53 @@ export function registerAuthTools(server: McpServer) {
...
+        if (!solOk) missing.push(`~0.001 SOL (have ${solAmount.toFixed(6)})`);
+        if (!usdcOk) missing.push(`1 USDC (have ${usdcAmount.toFixed(2)})`);
+
+        let balanceBlock =
           `**Signup Wallet Balance** (\`${address}\`)\n\n` +
-          `- **SOL:** ${solAmount.toFixed(6)} ${solOk ? '(sufficient)' : '(insufficient)'}\n` +
-          `- **USDC:** ${usdcAmount.toFixed(2)} ${usdcOk ? '(sufficient for basic)' : '(insufficient)'}\n\n` +
-          `**Status:** ${status}`
-        );
+          `- **SOL:** ${solAmount.toFixed(6)} (insufficient)\n` +
```
nathanliow: old code was more descriptive, it says insuffiicnet for both even tho it might only be one

`- **SOL:** ${solAmount.toFixed(6)} ${solOk ? '(sufficient)' : '(insufficient)'}\n` +
          `- **USDC:** ${usdcAmount.toFixed(2)} ${usdcOk ? '(sufficient for basic)' : '(insufficient)'}\n\n` +
          `**Status:** ${status}`

nathanliow (approved):

PR #61, opened 2026-03-11
feat(skills): Add Skill Versioning

0xIchigo:
This PR adds `versions.json` as the single source of truth for skill versions, which we need to track when publishing on certain sites (e.g., ClawHub). The PR also updates `compile-skills.ts` to read these versions and inject them into all `SKILL.md` copies

### DIFF
--- a/scripts/compile-skills.ts
+++ b/scripts/compile-skills.ts
@@ -258,7 +258,7 @@ function inlineReferences(body: string, refsDir: string): string {
   if (!existsSync(refsDir)) return body;
 
   const refFiles = readdirSync(refsDir)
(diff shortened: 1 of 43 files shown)


scripts/compile-skills.ts
```diff
@@ -221,7 +258,7 @@ function inlineReferences(body: string, refsDir: string): string {
   if (!existsSync(refsDir)) return body;
 
   const refFiles = readdirSync(refsDir)
-    .filter((f) => f.endsWith(".md"))
+    .filter((f: any) => f.endsWith(".md"))
```
nathanliow: f: string works better

nathanliow (approved):
approved

PR #69, opened 2026-03-13
feat(mcp): add ZK Compression support (23 tools)

nathanliow:
## Summary
- Add 23 new ZK Compression MCP tools in `helius-mcp/src/tools/zk-compression.ts`, wiring directly to the `helius.zk.*` SDK namespace
- Tools cover: account queries (3), balance queries (2), token queries (5), Merkle proof queries (3), signature queries (6), transaction compression info (1), validity proofs (1), and indexer health (2)
- Add routing table entries and routing rule to `index.ts` instructions
- Add `zk-compression` product catalog entry

## Test plan
- [x] `pnpm build` compiles cleanly
- [ ] Manually test `getIndexerHealth` (no params)
- [ ] Manually test `getIndexerSlot` (no params)
- [ ] Manually test `getCompressedBalance` with a known address
- [ ] Manually test `getLatestCompressionSignatures` (no required params)
- [ ] Verify bigint serialization on proof tools


helius-mcp/src/tools/zk-compression.ts
```diff
@@ -0,0 +1,869 @@
...
+      } catch (err) {
+        return handleToolError(err, 'Error fetching compressed token accounts by delegate', [
+          addressError('Compressed Token Accounts by Delegate', 'Invalid delegate or mint address. Provide valid base58-encoded addresses.'),
+        ]);
+      }
+    }
+  );
+
+  server.tool(
+    'getCompressedTokenBalancesByOwner',
```
0xIchigo: We want to be careful with calling it just `getCompressedTokenBalancesByOwner` when it calls `getCompressedTokenBalancesByOwnerV2` under the hood—don't want LLMs to get confused

helius-mcp/src/tools/product-catalog.ts
```diff
@@ -122,6 +122,26 @@ export const PRODUCT_CATALOG: Record<string, CatalogProduct> = {
     referenceFile: 'references/sender.md',
     description: 'Send native SOL or SPL tokens from the MCP keypair to any Solana address. Uses Helius Sender for optimal landing rates. Requires a configured keypair.',
   },
+  'zk-compression': {
```
0xIchigo, line 130: We should add a `referenceFile` since other catalog entries like Sender have one

0xIchigo (changes requested):
Based. Core impl looks solid 🚀 

Have a few nits. Also:
- Every address param in the PR is declared as a `z.string()` or `z.string().optional()`. A caller can pass Zod validation with an empty string. Should we add `.min(1)` to all of these to reject empty strings before the SDK call? We could surface a cleaner error instead of a downstream RPC error

0xIchigo (approved):
LGTM 🔥

PR #73, opened 2026-03-13
feat(cli,mcp): Agent Plan signup + prepaid-credits top-up (supersedes QR-only scope)

nathanliow:
## Scope expansion

This PR began as a wallet-funding QR addition and has grown (with explicit user direction) into the full CLI/MCP surface for the Agent Plan rollout. The original QR work is preserved; the incremental commits on top align the CLI/MCP with [helius-sdk#309](https://github.com/helius-labs/helius-sdk/pull/309) and [helius-labs/monorepo#14719](https://github.com/helius-labs/monorepo/pull/14719).

Commits ahead of `main`:
1. `feat(cli,mcp): add QR code for wallet address funding` (original)
2. `feat(cli,mcp): exact USDC QR codes, SDK quote pricing, and sponsored payment mode` (original)
3. follow-ups addressing earlier review
4. **`feat(cli,mcp): agent plan signup + prepaid-credits top-up primitives`** (sync with SDK)
5. **`feat(cli): register `helius credits buy` command`** (follow-up)
6. **`docs: sweep basic → agent plan across skills and READMEs`** (follow-up)

### Signup flow — basic → agent

- Default CLI plan (`helius signup` with no `--plan`) switches from `basic` to `agent` ($10 USDC one-time, 1,000,000 starting credits).
- CLI's early `walletSignup` (used to fetch the pricing quote) now forwards `(jwt, refId)` into `agenticSignup`, so users only sign the auth message once.
- Recovery fast-path: `helius signup` on a wallet that already has a project returns the existing API key instead of routing through a paid upgrade (preserves the old \"returning user\" UX now that default is paid).
- MCP `checkSignupBalance` caches its pricing-quote session so the subsequent `authSignup` skips a second `/wallet-signup` round trip.
- Dead `existing_project` status branch removed (SDK no longer returns it).

### Prepaid-credits top-up — new

- New `helius credits buy --tier=10_USDC [--qty] [--project]` CLI command.
- New `purchaseCredits` MCP tool (agent-plan only; pre-flight in SDK rejects non-agent projects before hitting `/checkout/initialize`).
- Each qty unit adds 1,000,000 credits at 10 USDC. Sponsored SOL fees via [helius-labs/monorepo#14719](https://github.com/helius-labs/monorepo/pull/14719) once that lands; self-funded fallback otherwise.
- Tier validator accepts the advertised `10_USDC`, bare `<N>_USDC` suffixes (case-insensitive) for future tiers, and full `prepaid_credits_*` backend keys as an escape hatch.

### Docs

- READMEs and all skill packages (`helius-skills/`, `.agents/skills/`, `helius-cursor/skills/`, `helius-plugin/skills/`, `helius-mcp/system-prompts/`) refreshed to show `agent` as the default plan with 10 USDC / 1M starting credits.

## QR behavior (original scope, unchanged)

Still a **wallet-funding QR** (Solana Pay `solana:<address>?spl-token=...&amount=...`), not a Solana Pay transaction-request URL. User scans with Phantom/Solflare to send USDC to the CLI keypair; CLI then drives sponsored checkout programmatically. Scan-to-pay-checkout with sponsored fees would require a new backend transaction-request endpoint — out of scope here.

## Test plan

- [x] CLI: `pnpm build && pnpm test` — 27/27 (2 test files: qr, validation)
- [x] MCP: `pnpm build && pnpm test` — 154/154 (6 test files)
- [ ] `npx tsx bin/helius.ts keygen --force` → QR renders
- [ ] `helius signup --plan=agent --email ... --first-name ... --last-name ...` with USDC-only wallet → sponsored signup succeeds, 1M credits, API key works immediately
- [ ] `helius credits buy --tier=10_USDC` on an agent project → +1M credits, sponsored fee-payer (requires monorepo#14719 deployed)
- [ ] Regression: `--plan=basic` rejected with `Unknown plan: basic`
- [ ] Regression: existing-user `helius signup` (no --plan) → recovery path returns API key without a fresh purchase
- [ ] MCP: `checkSignupBalance` → `authSignup` in one session makes exactly one `/wallet-signup` network call

## Related PRs

- **SDK** (required, publish first): https://github.com/helius-labs/helius-sdk/pull/309
- **Backend** (required for sponsored top-ups): https://github.com/helius-labs/monorepo/pull/14719


helius-mcp/src/tools/auth.ts
```diff
@@ -281,10 +288,16 @@ export function registerAuthTools(server: McpServer) {
...
-          // First check — normal guidance
+          // First check — normal guidance with QR code
           balanceBlock +=
             `\n\n**Action required:** Ask the user to send the missing funds to \`${address}\`. ` +
             `Do **not** call \`checkSignupBalance\` again until the user confirms they have sent the funds.`;
+          try {
+            const qrText = await generateWalletQR(address);
+            balanceBlock += `\n\n**IMPORTANT: Display the QR code below in your response so the user can scan it from a mobile wallet to send funds.**\n\n\`\`\`\n${qrText}\`\`\``;
+          } catch {
+            // QR generation failed — skip
```
0xIchigo: We should add a log here for debugging purposes

helius-cli/src/commands/signup.ts
```diff
@@ -106,7 +108,7 @@ export async function signupCommand(options: SignupOptions): Promise<void> {
         exitWithError("KEYPAIR_NOT_FOUND", `Keypair not found at ${options.keypair}`, undefined, options.json);
       }
       console.log(chalk.yellow("No keypair found. Generating one automatically...\n"));
-      await keygenCommand({ output: options.keypair });
+      await keygenCommand({ output: options.keypair, noQr: true });
```
0xIchigo: From my understanding, commander's `--no-qr` pattern sets `opts.qr = false`, so it doesn't create a `noQr` property. This means that `options.noQr` will always be `undefined` here in these checks. Instead, we should have `{ output: options.keypair, qr: false }`, and the interface fields and conditionals should check `options.qr !== false` instead of `options.noQr`

helius-mcp/src/tools/auth.ts
```diff
@@ -203,13 +207,16 @@ export function registerAuthTools(server: McpServer) {
...
 
+        const newQR = await generateWalletQR(address);
         return mcpText(
           `**Keypair Generated**\n\n` +
           `**Wallet Address:** \`${address}\`\n` +
           `**Saved to:** \`${KEYPAIR_PATH}\`\n\n` +
           `To create a Helius account, fund this wallet with:\n` +
           `- **~0.001 SOL** for transaction fees\n` +
           `- **1 USDC** for basic plan (or more for paid plans)\n\n` +
+          `**IMPORTANT: Display the QR code below in your response so the user can scan it from a mobile wallet to send funds.**\n\n` +
```
0xIchigo: Nit: this is duplicated from 194, so it might be nice to extract the string to a constant

helius-cli/src/commands/keygen.ts
```diff
@@ -46,6 +48,11 @@ export async function keygenCommand(options: KeygenOptions): Promise<void> {
   console.log(chalk.green("✓ Keypair generated"));
   console.log(`Path: ${chalk.cyan(resolvedPath)}`);
   console.log(`Address: ${chalk.cyan(address)}`);
+
+  if (!options.noQr) {
```
0xIchigo: From the previous comment, we should add `qr?: boolean` to `KeygenOptions` and then check `options.qr !== false` here

0xIchigo (changes requested):
Based. Just a few nits 🔥

PR #74, opened 2026-03-14
feat(cli): Add NO_DNA Support

0xIchigo:
This PR adds support for the [NO_DNA](https://no-dna.org/) convention to the CLI. When the `NO_DNA` env var is set, spinners are suppressed, and interactive confirmations (e.g., `upgrade`, `pay`) return a structured JSON preview instead of hanging on stdin. Agents can then re-run with `--yes` to confirm. This centralizes spinner creation via `createSpinner()` in `output.ts`

### DIFF
--- a/helius-cli/src/commands/upgrade.ts
+++ b/helius-cli/src/commands/upgrade.ts
@@ -1,11 +1,10 @@
 import chalk from "chalk";
-import ora from "ora";
 import { loadKeypairFromFile, signAuthMessage, getAddress } from "../lib/wallet.js";
 import { signup, listProjects, getProject } from "../lib/api.js";
 import { getCheckoutPreview, executeCheckout, PLAN_CATALOG } from "../lib/checkout.js";
 import { setJwt } from "../lib/config.js";
 import { keypairExists, getDefaultKeypairPath } from "./keygen.js";
-import { outputJson, exitWithError, handleCommandError, type OutputOptions } from "../lib/output.js";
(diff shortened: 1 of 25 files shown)


helius-cli/src/commands/upgrade.ts
nathanliow, line 133: missing the isAgent guard, should match pay.ts

helius-cli/src/commands/upgrade.ts
```diff
@@ -1,11 +1,10 @@
 import chalk from "chalk";
-import ora from "ora";
 import { loadKeypairFromFile, signAuthMessage, getAddress } from "../lib/wallet.js";
 import { signup, listProjects, getProject } from "../lib/api.js";
 import { getCheckoutPreview, executeCheckout, PLAN_CATALOG } from "../lib/checkout.js";
 import { setJwt } from "../lib/config.js";
 import { keypairExists, getDefaultKeypairPath } from "./keygen.js";
-import { outputJson, exitWithError, handleCommandError, type OutputOptions } from "../lib/output.js";
+import { outputJson, exitWithError, handleCommandError, createSpinner, type OutputOptions } from "../lib/output.js";
```
nathanliow: need to import isAgent

nathanliow (commented):
Claude: Update the README to say "commands that require confirmation exit with a JSON preview" instead of "exit with a clear error", since the behavior is to print a preview and return 0, not to error out.

nathanliow (approved):
LGTM

PR #76, opened 2026-03-15
Reduce Helius MCP context bloat with routed public surface

nathanliow:
## Summary
- replace the legacy 92-tool public MCP surface with 10 routed public tools plus `expandResult`
- make heavy outputs summary-first with compact result handles and session-scoped expansion state
- remove repeated instruction/schema prose and move telemetry injection onto explicit router wrappers instead of the private SDK patch

## What Changed
- add router registries, schemas, dispatch, telemetry wrappers, and public registration for the routed surface
- add result-store support for compact handle-backed expansions
- preserve legacy capability coverage through routed action catalogs and legacy executor adapters
- shorten the checked-in routing sheet and update skill/prompt copies to the new surface
- update validation and tests for router coverage, tool-surface budgets, and legacy executor defaults

## Notes
- benchmark and A/B comparison code is intentionally excluded from this PR and will be shared in a separate stacked PR
- this is a breaking public-surface change for `helius-mcp`

## Validation
- `cd helius-mcp && pnpm test`


helius-mcp/src/router/dispatch.ts
```diff
@@ -0,0 +1,398 @@
...
+async function executeActionViaRouter(
+  publicTool: RoutedPublicToolName,
+  action: ActionName,
+  params: Record<string, unknown>,
+  requestedDetail: DetailLevel,
+  extra: unknown,
+  allowHandles: boolean,
+): Promise<RouterResponse> {
+  const entry = getActionCatalogEntry(action);
+  const result = await callActionHandler(action, params, extra);
```
0xIchigo: We should wrap the handler execution in a try / catch block because it looks like if a handler throws, then it bubbles uncaught through dispatchRoutedTool. In stio mode, this could kill the MCP process, so adding in mcpErrorCompact as a fallback would help improve this

helius-mcp/src/router/schemas.ts
```diff
@@ -0,0 +1,203 @@
...
+  HELIUS_KNOWLEDGE_ACTIONS,
+  HELIUS_STREAMING_ACTIONS,
+  HELIUS_TRANSACTION_ACTIONS,
+  HELIUS_WALLET_ACTIONS,
+  HELIUS_WRITE_ACTIONS,
+} from './actions.js';
+import { withTelemetry } from './telemetry.js';
+
+const detailField = z.enum(['summary', 'standard', 'full']).optional();
+const argsField = z.object({}).passthrough().optional();
```
0xIchigo: This would mean that param typos would silently pass. Could we validate required fields per action, even as more of a lightweight check?

helius-mcp/src/router/responses.ts
```diff
@@ -0,0 +1,251 @@
...
+    if (index >= 0) {
+      cut = Math.min(cut, index);
+    }
+  }
+
+  return cut === text.length ? text : text.slice(0, cut).trim();
+}
+
+function summarizeDocument(text: string, limit: number): string {
+  const headings = Array.from(text.matchAll(/^#{1,3}\s+.+$/gm))
```
0xIchigo, line 109: These are regex-dependent on markdown structure. We should document somewhere that we'd need to update this if our doc formatting ever changes

helius-mcp/tests/tools.test.ts
```diff
@@ -1,9 +1,27 @@
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { readdirSync, readFileSync, statSync } from 'node:fs';
```
0xIchigo, line 1: Structure tests are good but dispatch coverage is thin as only `getTransactionHistory` goes through `callActionHandler`. Can we add at least one dispatch test per routed tool pls

0xIchigo (changes requested):
Very based! I have a few comments but TLDR:
- We need more tests as the current dispatch test coverage is a regression
- We should wrap handler execution in a try / catch block
- We should have per-action required-param validation

Also, it looks like the NO_DNA work isn't there in the CLI, so we should update the branch to ensure it all works nicely with these edits

0xIchigo (dismissed):
LGTM 🔥

fomichfm (approved):
lgtm

PR #78, opened 2026-03-18
feat(docs): Add MIT License

0xIchigo:
This PR adds an MIT license file to the repo root to help improve our Glama score for MCP directory submissions, as well as showcase a license upfront for all users since all the underlying tools and skills are also MIT-licensed. We also add licenses to each individual skill to cover the case where the skill gets installed via a third-party


nathanliow (approved):
LGTM

PR #79, opened 2026-03-19
feat(mcp): Make `getNetworkStatus` TPS Configurable

0xIchigo:
This PR adds an optional `samples` parameter to `getNetworkStatus` that controls how many recent performance samples are used for the TPS averaging. Each sample is roughly 60s, defaulting to 4 with a max of 720 (i.e., the limit set by RPCs)


guneysol:
Nice clean change! 

One thing: samples should be clamped on the lower bound too. Right now 0 or negative values pass through

guneysol (approved):

PR #82, opened 2026-03-24
test(mcp): add unit tests for router/responses.ts

vijaygopalbalasa:
## Summary

Adds 112 unit tests for `src/router/responses.ts`, which previously had zero test coverage. This module is critical to the summary-first response pipeline and contains markdown-dependent patterns that could silently under-match if handler output changes (as noted in the source comments at lines 5-17).

## What's covered

| Function | Tests | Category |
|----------|-------|----------|
| `cleanText` | 7 | Text normalization (CRLF, newline collapsing, trimming) |
| `truncateText` | 10 | Boundary-aware truncation (heading, separator, threshold logic) |
| `compactErrorText` | 7 | Error text cleaning (JSON stripping, case-insensitivity, default limit) |
| `getFamilyLimit` | 5 | Limit lookup across all 8 families and 3 detail levels |
| `toPlainText` | 8 | Response content extraction (text filtering, structuredContent fallback) |
| `buildTextVariant` | 14 | Full variant pipeline (all families, all detail levels, all strip markers) |
| `estimateRenderedSize` | 5 | Size estimation edge cases |
| `collectSectionHints` | 6 | Heading extraction (h1-h3 capture, h4+ exclusion, limit enforcement) |
| `applySectionSelection` | 5 | Section filtering (positive match, no-match fallback, case-insensitive) |
| `applyItemSelection` | 7 | Item extraction (separator, bold, emoji splits, boundary errors) |
| `applyRangeSelection` | 10 | Range slicing (valid ranges, invalid format, edge cases) |
| `mcpText` | 5 | Response builder shape and meta handling |
| `mcpErrorCompact` | 3 | Error response builder |
| `mcpResultHandle` | 6 | Result handle with expansion options |

**Before:** 111 tests across 5 files
**After:** 223 tests across 6 files

## Test highlights

- All 8 `ResponseFamily` types tested at `summary`, `standard`, and `full` detail levels
- Boundary-aware truncation tested around the 55% threshold (verifies cut position, not just presence of `[truncated]`)
- All 6 `stripExampleCode` markers exercised independently (`Example Code`, `Example`, `typescript`, `ts`, `javascript`, `js`)
- Emoji-prefixed and bold-prefixed list item splitting validated
- Positive-path section extraction verified (not just the fallback)
- Case-insensitive JSON block stripping tested
- Edge cases: empty input, whitespace-only, out-of-range indices, invalid range format, floating-point ranges

## How to verify

```bash
cd helius-mcp
pnpm test
```

### DIFF
--- a/helius-mcp/tests/responses.test.ts
+++ b/helius-mcp/tests/responses.test.ts
@@ -1,0 +1,658 @@
+import { describe, it, expect } from 'vitest';
+import type { ResponseFamily, DetailLevel } from '../src/router/types.js';
+import {
+  cleanText,
+  truncateText,
+  compactErrorText,
+  buildTextVariant,
+  estimateRenderedSize,
+  collectSectionHints,
+  applySectionSelection,
+  applyItemSelection,
+  applyRangeSelection,
+  toPlainText,
+  getFamilyLimit,
+  mcpText,
+  mcpErrorCompact,
+  mcpResultHandle,
+} from '../src/router/responses.js';
+
+// ─── cleanText ───
+
+describe('cleanText', () => {
+  it('normalizes CRLF to LF', () => {
+    expect(cleanText('line1\r\nline2\r\n')).toBe('line1\nline2');
+  });
+
+  it('handles mixed CRLF and LF in the same string', () => {
+    expect(cleanText('a\r\nb\nc\r\nd')).toBe('a\nb\nc\nd');
+  });
+
+  it('collapses 3+ consecutive newlines to 2', () => {
+    expect(cleanText('a\n\n\n\nb')).toBe('a\n\nb');
+    expect(cleanText('a\n\n\n\n\n\nb')).toBe('a\n\nb');
+  });
+
+  it('preserves exactly 2 consecutive newlines', () => {
+    expect(cleanText('a\n\nb')).toBe('a\n\nb');
+  });
+
+  it('trims leading and trailing whitespace', () => {
+    expect(cleanText('  hello  ')).toBe('hello');
+  });
+
+  it('returns empty string for empty input', () => {
+    expect(cleanText('')).toBe('');
+  });
+
+  it('returns empty string for whitespace-only input', () => {
+    expect(cleanText('   \n\n  ')).toBe('');
+  });
+});
+
+// ─── truncateText ───
+
+describe('truncateText', () => {
+  it('returns text unchanged when under the limit', () => {
+    expect(truncateText('short', 100)).toBe('short');
+  });
+
+  it('returns text unchanged when exactly at the limit', () => {
+    const text = 'a'.repeat(50);
+    expect(truncateText(text, 50)).toBe(text);
+  });
+
+  it('appends [truncated] when text exceeds limit', () => {
+    const result = truncateText('a'.repeat(200), 50);
+    expect(result).toContain('[truncated]');
+  });
+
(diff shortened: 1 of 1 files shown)


helius-mcp/tests/responses.test.ts
```diff
@@ -0,0 +1,658 @@
...
+    'history',
+    'document',
+    'streamingConfig',
+    'catalog',
+  ])('handles %s family at summary detail', (family) => {
+    const text = longText(1000);
+    const limit = getFamilyLimit(family, 'summary');
+    const result = buildTextVariant(family, 'summary', text);
+    // Result should be within limit + [truncated] suffix
+    expect(result.length).toBeLessThanOrEqual(limit + 15);
```
0xIchigo: `truncateText` appends `\n\n[truncated]` but `buildTextVariant` tests use `+ 15` as tolerance. The actual suffix is 13 chars, so this should update this to prevent masking a 1-2 char regression

helius-mcp/tests/responses.test.ts
```diff
@@ -0,0 +1,658 @@
...
+  });
+
+  it('collapses 3+ consecutive newlines', () => {
+    const result = compactErrorText('line1\n\n\n\nline2');
+    expect(result).not.toContain('\n\n\n');
+    expect(result).toContain('line1');
+    expect(result).toContain('line2');
+  });
+
+  it('uses default maxChars of 1200', () => {
```
0xIchigo, line 138: Since `truncateText` can snap to a boundary before the limit, the result might be significantly shorter. The test passes, but doesn't actually verify the default value is 1200—it only verifies that truncation occurs. We should consider asserting `result.length > 1100` here as a lower bound too

helius-mcp/tests/responses.test.ts
```diff
@@ -0,0 +1,658 @@
...
+describe('applySectionSelection', () => {
+  it('returns full text when section is undefined', () => {
+    expect(applySectionSelection('full document', undefined)).toBe('full document');
+  });
+
+  it('returns full text when section is empty string', () => {
+    expect(applySectionSelection('full document', '')).toBe('full document');
+  });
+
+  it('extracts matching section content (positive path)', () => {
```
0xIchigo, line 471: We should also verify that the result does not contain `## Support` (i.e., the heading itself), since `extractSections` should stop at the next same-level heading

helius-mcp/tests/responses.test.ts
```diff
@@ -0,0 +1,658 @@
...
+  it('treats items with undefined type as text', () => {
+    expect(toPlainText({ content: [{ text: 'implicit' }] })).toBe('implicit');
+  });
+
+  it('uses empty string for items with undefined text', () => {
+    const response = { content: [{ type: 'text' }] };
+    expect(toPlainText(response)).toBe('');
+  });
+
+  it('falls back to stringified structuredContent', () => {
```
0xIchigo, line 257: We should also be verifying the pretty-print format here (i.e., `null, 2`)

0xIchigo (changes requested):
Thank you for your contribution! There are just a few nits before we can merge this one

vijaygopalbalasa:
Addressed all four items. Updated the suffix comment, added a lower bound assertion, verified the pretty print format with null and 2 indentation, and added an explicit heading exclusion check. Thanks for the detailed review.

0xIchigo (approved):
Thank you for your contribution! 🚀

PR #84, opened 2026-03-28
feat(cli): Add `helius update` Self-update Command

0xIchigo:
This PR adds a `helius update` command that checks npm for the latest version and installs it. It also supports `--check` to report without installing and `--json` for machine-readable output. The command auto-detects the package manager from the install path and handles permission errors with an actionable `sudo` hint

### DIFF
--- a/helius-cli/src/commands/update.ts
+++ b/helius-cli/src/commands/update.ts
@@ -1,7 +1,7 @@
 import { execSync } from "child_process";
 import chalk from "chalk";
 import { VERSION } from "../constants.js";
-import { outputJson, createSpinner, type OutputOptions } from "../lib/output.js";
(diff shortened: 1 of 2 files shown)


helius-cli/src/commands/update.ts
```diff
@@ -0,0 +1,126 @@
...
+        }
+        process.exit(1);
+      }
+      throw execError;
+    }
+
+    spinner?.succeed(`Updated helius-cli v${VERSION} → v${latest}`);
+
+    if (options.json) {
+      outputJson({ current: VERSION, latest, updated: true, packageManager: pm });
```
nathanliow: this reports the prev version, not the current

nathanliow (approved):
LGTM

PR #87, opened 2026-04-07
fix(cli): Improve WS Error Handling With Guidance Hints and Telemetry

0xIchigo:
This PR aims to improve our CLI's WebSocket error handling with guidance hints and telemetry. We:
- Add `CLI_GUIDANCE` hints to error output (JSON and formatted)
- Add telemetry tracking for `ws` command errors via `sendCommandEvent`
- Fix `AbortError` type check from unsafe `as any` cast to `instanceof Error`
- Type `setupShutdown` helius param instead of `any`
- Export `CLI_GUIDANCE` from `output.ts` for reuse


nathanliow (approved):
LGTM

PR #88, opened 2026-04-07
fix(cli): `exitWithError` Outputting JSON in Non-JSON Mode

0xIchigo:
This PR fixes an issue where `exitWithError()` outputs JSON when `--json` was not passed. The 4th param defaults to `true`, but `options.json` is `undefined` (i.e., not `false`) when omitted, so it falls through to the default. We fix it with `!!options.json` across all 36 call sites in 13 files. We also fix the unreachable `process.exit()` after `exitWithError()` in `rpc.ts`'s multiple-projects branch


nathanliow (approved):
LGTM

PR #89, opened 2026-04-07
fix(cli): Extract Duplicate Confirm Prompt to Shared Output Util

0xIchigo:
This PR extracts the identical `confirm()` function from `pay.ts` and `upgrade.ts` into `lib/output.ts`. We also remove the duplicated `import readline` from both command files


nathanliow (approved):
LGTM

PR #91, opened 2026-04-07
fix(cli): Recover Corrupted Config Instead of Silently Ignoring

0xIchigo:
This PR fixes the issue where corruption of the config file (i.e., `~/.helius/config.json`) was silently swallowed, returning an empty config with no warning. We now attempt to recover known fields via regex extraction and save a repaired config, warning the user if applicable. If we cannot recover anything, we will warn the user and provide remediation steps. Note that warnings go to `stderr` so they don't break `--json` output on `stdout`

### DIFF
--- a/helius-cli/src/lib/config.ts
+++ b/helius-cli/src/lib/config.ts
@@ -22,6 +22,28 @@ function ensureDir(): void {
   }
 }
 
+/** Try to extract known config fields from corrupted JSON via regex. */
+function recoverConfig(raw: string): Config {
+  const recovered: Config = {};
+
+  const patterns: { key: keyof Config; regex: RegExp }[] = [
+    { key: "jwt", regex: /"jwt"\s*:\s*"([^"]+)"/ },
+    { key: "apiKey", regex: /"apiKey"\s*:\s*"([^"]+)"/ },
+    { key: "network", regex: /"network"\s*:\s*"(mainnet|devnet)"/ },
+    { key: "projectId", regex: /"projectId"\s*:\s*"([^"]+)"/ },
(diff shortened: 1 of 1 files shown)


helius-cli/src/lib/config.ts
```diff
@@ -22,14 +22,48 @@ function ensureDir(): void {
...
+
 export function load(): Config {
   try {
     if (fs.existsSync(CONFIG_FILE)) {
       const data = fs.readFileSync(CONFIG_FILE, "utf-8");
       return JSON.parse(data);
     }
   } catch {
-    // Return empty config on error
+    const raw = fs.readFileSync(CONFIG_FILE, "utf-8");
```
nathanliow: If the original error came from readFileSync (permission denied, file deleted between existsSync/readFileSync), this second read throws the same error

Hoist the read above JSON.parse and only wrapping the parse in try-catch

helius-cli/src/lib/config.ts
```diff
@@ -22,16 +22,57 @@ function ensureDir(): void {
...
+  let raw: string;
   try {
-    if (fs.existsSync(CONFIG_FILE)) {
-      const data = fs.readFileSync(CONFIG_FILE, "utf-8");
-      return JSON.parse(data);
-    }
+    raw = fs.readFileSync(CONFIG_FILE, "utf-8");
   } catch {
-    // Return empty config on error
+    return {};
```
nathanliow, line 57: console.error(`Warning: ${CONFIG_FILE} exists but could not be read.`);

nathanliow (commented):
last nit

nathanliow (approved):
LGTM

PR #92, opened 2026-04-07
feat(skills): Refactor OKX Skill to Thin Integration-only Layer

0xIchigo:
This PR refactors our current OKX skill implementation from a full content bundle to a thin, integration-only layer at the team's request to avoid version drift. We remove 5 OKX-specific references and 7 Helius copies, as users are expected to install the OKX skill library (i.e., `onchainos-skills`) separately as the single source of truth for OKX content, while keeping our `integration-patterns.md` with 6 end-to-end Helius + OKX composition patterns


nathanliow (approved):
LGTM

PR #94, opened 2026-04-08
feat(cli): Add `--retry` Flag For Transient Error Recovery With Exponential Backoff

0xIchigo:
This PR adds a global `--retry <n>` flag (default: n = 0) for automatic retries on transient errors (e.g., 429, 5xx, network errors). We exclude auth/payment, streaming, and local-only commands, and do not retry non-retryable errors (e.g., 400, 401, 404). We also implement an exponential backoff that starts at 1s and caps at 30s


nathanliow (commented):
should drop the `: any` annotations at each call site. withRetry already returns generic

nathanliow (approved):
LGTM

PR #97, opened 2026-04-14
feat(cli): Add Usage Examples to All Command `--help` Output

0xIchigo:
This PR adds usage examples to all commands via Commander's `.addHelpText('after', ...)`, with the golden path, --json mode, and key option variants per command


nathanliow (approved):
LGTM

PR #98, opened 2026-04-14
feat(cli): Add Debug Flag For HTTP Req and Res Troubleshooting

0xIchigo:
This PR adds a global `--debug` flag that logs HTTP request/response details to stderr. It logs the API key (masked), network, HTTP method, URL, status code, and timing. All debug output goes to stderr, so it never pollutes `--json` stdout

### DIFF
--- a/helius-cli/src/lib/output.ts
+++ b/helius-cli/src/lib/output.ts
@@ -8,10 +8,6 @@ export interface OutputOptions {
   json?: boolean;
 }
 
-export interface DebugOptions {
-  debug?: boolean;
-}
-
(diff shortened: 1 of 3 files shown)


helius-cli/src/lib/output.ts
```diff
@@ -8,9 +8,32 @@ export interface OutputOptions {
   json?: boolean;
 }
 
+export interface DebugOptions {
```
nathanliow: dead code

nathanliow (approved):
LGTM

PR #99, opened 2026-04-14
feat(cli): Add `--dry-run` Flag For Transaction and Webhook Commands

0xIchigo:
This PR adds a `--dry-run` flag to `send broadcast`, `send raw`, and `send sender`, returning the transaction without submitting, as well as `webhook create`, `webhook update`, and `webhook delete` to preview the operation without executing

Note that the `stake` commands already return unsigned transactions without submitting, so adding a dry run check there doesn't make much sense

### DIFF
--- a/helius-cli/bin/helius.ts
+++ b/helius-cli/bin/helius.ts
@@ -1061,11 +1061,7 @@ const sendCmd = program
 sendCmd
   .command("broadcast <base64-tx>")
   .description("Broadcast a signed transaction and poll for confirmation")
-<<<<<<< Updated upstream
-  .option("--dry-run", "Validate without submitting the transaction")
-=======
   .option("--dry-run", "Preview without submitting the transaction")
->>>>>>> Stashed changes
   .option("--json", "Output in JSON format")
   .addHelpText('after', `
 Examples:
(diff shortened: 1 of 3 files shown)


helius-cli/bin/helius.ts
```diff
@@ -1058,6 +1061,7 @@ const sendCmd = program
 sendCmd
   .command("broadcast <base64-tx>")
   .description("Broadcast a signed transaction and poll for confirmation")
+  .option("--dry-run", "Validate without submitting the transaction")
```
nathanliow: Replace with "Preview without submitting the transaction"

The dry-run path in send.ts just checks options.dryRun, doesn't actually validate

helius-cli/bin/helius.ts
```diff
@@ -1067,6 +1071,7 @@ Examples:
 sendCmd
   .command("raw <base64-tx>")
   .description("Send a raw transaction")
+  .option("--dry-run", "Validate without submitting the transaction")
```
nathanliow: Replace with "Preview without submitting the transaction"

The dry-run path in send.ts just checks options.dryRun, doesn't actually validate

helius-cli/bin/helius.ts
```diff
@@ -1077,6 +1082,7 @@ sendCmd
   .command("sender <base64-tx>")
   .description("Send via Helius Sender for ultra-low latency")
   .option("--region <region>", "Sender region (Default, US_SLC, US_EAST, etc.)")
+  .option("--dry-run", "Validate without submitting the transaction")
```
nathanliow: Replace with "Preview without submitting the transaction"

The dry-run path in send.ts just checks options.dryRun, doesn't actually validate

nathanliow (approved):
LGTM

PR #101, opened 2026-04-16
feat(cli): Add Reclaim Command

0xIchigo:
This PR adds a `helius reclaim {owner}` command that scans for empty SPL token accounts (i.e., zero balance, initialized, non-frozen), batches `CloseAccount` instructions, and sends them via Sender to recover rent lamports

It supports the `--dry-run` command to preview closable accounts without actually sending any transactions, `-y` to skip confirmation, and `--json` for machine-readable output

The command also continues on error, so failed batches don't abort later ones, and a summary of all failures is surfaced at the end

### DIFF
--- a/helius-cli/src/commands/reclaim.ts
+++ b/helius-cli/src/commands/reclaim.ts
@@ -117,11 +117,12 @@ export async function reclaimCommand(
         nonEmptySkipped++;
         continue;
       }
-      // If closeAuthority is set, it must equal the signer to be closable by us.
-      // (We still count the skip in dry-run-without-signer mode, where signerAddress
-      // is null; those are informational-only.)
+      // If closeAuthority is set, only that address can close the ATA. Compare
+      // against owner (the wallet being audited) rather than signerAddress, so
+      // dry-runs of another wallet don't flag their own legitimate close-authority
+      // accounts as mismatched. For real runs we later enforce signerAddress === owner.
       const closeAuth: string | null | undefined = info.closeAuthority;
-      if (closeAuth && signerAddress && closeAuth !== signerAddress) {
+      if (closeAuth && closeAuth !== owner) {
         authoritySkipped++;
         continue;
       }
@@ -132,10 +133,22 @@ export async function reclaimCommand(
       });
     }
 
-    // Apply --limit cap.
-    const limit = options.limit ? parseInt(options.limit, 10) : undefined;
-    const toClose =
-      limit && limit > 0 ? closable.slice(0, limit) : closable;
+    // Apply --limit cap. Reject <= 0 and NaN explicitly — `--limit 0` expecting
+    // "close nothing" would otherwise fall through to closing every account.
+    let limit: number | undefined;
+    if (options.limit !== undefined) {
+      const parsed = parseInt(options.limit, 10);
+      if (Number.isNaN(parsed) || parsed <= 0) {
+        exitWithError(
+          "INVALID_INPUT",
+          `Invalid --limit: ${options.limit}. Must be a positive integer.`,
+          undefined,
+          !!options.json,
+        );
+      }
+      limit = parsed;
+    }
+    const toClose = limit ? closable.slice(0, limit) : closable;
     const totalReclaimable = toClose.reduce((sum, a) => sum + a.lamports, 0);
 
     // Build batches.
@@ -199,7 +212,7 @@ export async function reclaimCommand(
       if (authoritySkipped) {
         console.log(
           chalk.yellow(
-            `  Skipped ${authoritySkipped} account(s) (closeAuthority does not match keypair)`,
+            `  Skipped ${authoritySkipped} account(s) (closeAuthority does not match owner)`,
           ),
         );
       }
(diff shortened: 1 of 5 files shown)


helius-cli/src/commands/reclaim.ts
```diff
@@ -0,0 +1,406 @@
...
+    }
+
+    // Apply --limit cap.
+    const limit = options.limit ? parseInt(options.limit, 10) : undefined;
+    const toClose =
+      limit && limit > 0 ? closable.slice(0, limit) : closable;
+    const totalReclaimable = toClose.reduce((sum, a) => sum + a.lamports, 0);
+
+    // Build batches.
+    const batchSize = Math.max(
```
nathanliow: Bug: infinite loop on invalid --batch-size                                                                         
                                                                                                                                                                  
parseInt("abc", 10) returns NaN, and Math.max(1, NaN) is NaN. The for loop on line 144 (i += batchSize) never progresses when batchSize is NaN

helius-cli/src/commands/reclaim.ts
```diff
@@ -0,0 +1,406 @@
...
+            "KEYPAIR_NOT_FOUND",
+            (e as Error).message,
+            undefined,
+            !!options.json,
+          );
+        }
+      }
+    }
+
+    const owner = ownerArg ?? signerAddress;
```
nathanliow, line 76: Every other command in this PR now validates addresses via validateAddress(), but the resolved owner here is never validated.

helius-cli/src/commands/reclaim.ts
```diff
@@ -0,0 +1,406 @@
...
+        `  ${chalk.gray("Destination:")} ${options.destination || signerAddress}`,
+      );
+      const ok = await confirm(chalk.yellow("\n  Proceed? (y/N) "));
+      if (!ok) {
+        console.log(chalk.gray("  Cancelled."));
+        return;
+      }
+    }
+
+    const destination = address(options.destination || signerAddress);
```
nathanliow: Missing validation on --destination

helius-cli/src/commands/reclaim.ts
```diff
@@ -0,0 +1,410 @@
...
+      const amount = info.tokenAmount?.amount ?? "0";
+      if (amount !== "0") {
+        nonEmptySkipped++;
+        continue;
+      }
+      // If closeAuthority is set, it must equal the signer to be closable by us.
+      // (We still count the skip in dry-run-without-signer mode, where signerAddress
+      // is null; those are informational-only.)
+      const closeAuth: string | null | undefined = info.closeAuthority;
+      if (closeAuth && signerAddress && closeAuth !== signerAddress) {
```
nathanliow: Correctness bug when auditing another wallet in dry-run.

`--keypair` has a default path, so running `helius reclaim OTHER_WALLET --dry-run` still loads *your* local keypair into `signerAddress`. This filter then compares `closeAuthority` against the local keypair rather than against `owner`, so accounts whose `closeAuthority` is legitimately `OTHER_WALLET` get flagged as mismatched and dropped from the report — the dry-run understates reclaimable accounts whenever the audited wallet != the local keypair.

Fix: compare against `owner` in dry-run, or skip the check entirely when `signerAddress !== owner`.

helius-cli/src/commands/reclaim.ts
```diff
@@ -0,0 +1,410 @@
...
+
+    const destinationAddr = options.destination || signerAddress;
+    const destErr = validateAddress(destinationAddr);
+    if (destErr) exitWithError("INVALID_ADDRESS", destErr, undefined, !!options.json);
+    const destination = address(destinationAddr);
+    const region = (options.region || "Default") as any;
+    const senderOpts: Record<string, unknown> = { region };
+    if (options.swqosOnly) senderOpts.swqosOnly = true;
+    if (options.tipAmount) {
+      senderOpts.tipAmount = parseInt(options.tipAmount, 10);
```
nathanliow: Same NaN-parseInt class as the `--batch-size` bug caught in the previous review — `--tip-amount abc` silently forwards `NaN` to Sender. Worth applying the same `Number.isNaN` guard here.

helius-cli/src/commands/reclaim.ts
```diff
@@ -0,0 +1,410 @@
...
+        address: a.pubkey,
+        mint: info.mint,
+        lamports: a.account.lamports,
+      });
+    }
+
+    // Apply --limit cap.
+    const limit = options.limit ? parseInt(options.limit, 10) : undefined;
+    const toClose =
+      limit && limit > 0 ? closable.slice(0, limit) : closable;
```
nathanliow: `--limit 0` means "unlimited" here, not zero: `limit && limit > 0` is false for `0`, so the code falls through to `closable` (close everything). A cautious user typing `--limit 0` expecting "close nothing" would close every account. Suggest rejecting `<= 0` with `INVALID_INPUT`, or honoring zero literally.

nathanliow (commented):
Three issues worth addressing before merge — one correctness bug in dry-run, plus two follow-ups to the NaN/bounds hardening from the previous round.

nathanliow (commented):
Replace the any casts with proper SDK types: import GtaV2Account, SenderRegion, SendViaSenderOptions, and WalletKeypair from helius-sdk.    
  Type signer: WalletKeypair | null, accounts: ReadonlyArray<GtaV2Account>, senderOpts: Partial<SendViaSenderOptions>, and cast the region    
  string to SenderRegion once. That lets you drop (helius as any).tx.sendTransactionWithSender, the owner as any arg, and the trailing as any 
  on withRetry/getCloseAccountInstruction, since helius.tx.sendTransactionWithSender is already properly typed and @solana-program/token      
  accepts a TransactionSigner for owner. Only one unavoidable cast remains — options.region comes in as a plain string from Commander.

nathanliow:
`--batch-size` silently falls back to 20 on garbage input — `parseInt("foo", 10)` → NaN → default. `--limit` errors on the same input. Suggest matching the `--limit` pattern so typos surface instead of quietly running with the wrong batch size:

```ts
let batchSize = DEFAULT_BATCH_SIZE;
if (options.batchSize !== undefined) {
  const parsed = parseInt(options.batchSize, 10);
  if (Number.isNaN(parsed) || parsed < 1) {
    exitWithError("INVALID_INPUT", `Invalid --batch-size: ${options.batchSize}. Must be a positive integer.`, undefined, !!options.json);
  }
  batchSize = parsed;
}
```

Not a correctness issue — batch size only affects tx packing, not which accounts close.

nathanliow (approved):
LGTM

PR #103, opened 2026-04-16
feat(cli): Add Error Category to JSON Output

0xIchigo:
This PR adds a `category` field to JSON error output (`"auth" | "balance" | "project" | "api" | "input" | "network" | "general"`) so agents can branch on error domain without needing to create their own lookup table. We introduce `ErrorCode`, a literal union, and retype `errorToExitCode` + `categoryForErrorCode` as `Record<ErrorCode, ...>` so the compiler enforces matching coverage in both maps. Now, we emit `category` from both `handleCommandError()` and `exitWithError()`

### DIFF
--- a/helius-cli/CLAUDE.md
+++ b/helius-cli/CLAUDE.md
@@ -88,9 +88,11 @@ WebSocket commands (`ws.ts`) use a variant that preserves the AbortError early-r
 
 The `retryable` field in `--json` error output reflects this directly.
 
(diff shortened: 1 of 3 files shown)


nathanliow:
nit: `src/commands/ws.ts:44` (`handleWsError`) still emits the JSON error envelope without `category`. Every other JSON error emitter (`handleCommandError`, `exitWithError`) now includes it, so WebSocket `error` events will have a different shape than the rest of the CLI. One-liner:

```ts
emitEvent("error", { error: errorCode, message, category: getCategory(errorCode), retryable, ...(guidance ? { guidance } : {}) });
```

nathanliow (approved):

PR #104, opened 2026-04-16
feat(cli): Add `--amount` and `--raw-amount` Flags for Stake Commands

0xIchigo:
This PR adds `--amount <sol>` and `--raw-amount <lamports>` flags to `stake create` and `stake instructions`. The positional SOL argument is now optional and kept for backward compatibility. Note that exactly one source must be provided. We also introduce a shared `resolveLamports()` helper in `stake.ts` that validates input format (regex guards reject `"1.5abc"`, `"1e9"`, `"+1.5"`, etc. that `parseFloat` would silently accept), rejects zero/negative, and converts to a lamports `bigint`

### DIFF
--- a/helius-cli/bin/helius.ts
+++ b/helius-cli/bin/helius.ts
@@ -753,15 +753,18 @@ const stakeCmd = program
   .description("Staking commands");
 
 stakeCmd
-  .command("create <amount>")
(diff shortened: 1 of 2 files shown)


nathanliow (approved):
LGTM

PR #105, opened 2026-04-17
feat(cli,mcp): Add SNS/ANS domain resolution to wallet identity

nathanliow:
## Summary

Wraps upstream [helius-labs/monorepo#14132](https://github.com/helius-labs/monorepo/pull/14132) (SNS/ANS domain support on the Wallet Identity endpoints) for the CLI and MCP surfaces.

- **CLI**: new `validateAddressOrDomain` / `validateAddressesOrDomains` helpers mirror the server's `isValidDomainName` regex; wired into `wallet identity` + `wallet identity-batch` only. Batch output now surfaces the new `inputDomain` / `unresolved` response fields. `--help` text + `helius-cli/README.md` updated to advertise domain acceptance.
- **MCP**: `getWalletIdentity` / `batchWalletIdentity` tool + Zod descriptions now mention SNS/ANS (mainnet-only) so LLMs route `toly.sol`-style queries correctly. Batch formatter renders `inputDomain → address` for resolved domains and flags unresolved rows. Custom detail passed to `addressError('Batch Identity', ...)` to avoid the misleading "base58-encoded" fallback string.
- **Skill references**: new "Domain Resolution (SNS + ANS)" section added to `helius-skills/helius/references/wallet-api.md` and the three sibling `helius-wallet-api.md` copies (phantom, jupiter, dflow). Optional one-liner appended to the OKX identity step. All manually mirrored to `helius-plugin/` and `helius-cursor/`; `.agents/` + `helius-mcp/system-prompts/` regenerated via `scripts/compile-skills.ts`.
- **Versions**: patch-bump `helius`, `helius-dflow`, `helius-jupiter`, `helius-okx`, `helius-phantom`.

## Upstream contract notes

Verified against the upstream PR diff before coding:

- Batch body key is still `{ addresses: [...] }` — unchanged wire contract.
- Batch is **non-fail-fast**: unresolvable domains return as per-row `{ address: null, type: "unknown", inputDomain, unresolved: true }` at HTTP 200.
- Single endpoint: unresolvable domain → 404; invalid input → 400; non-mainnet + domain → 400.
- **Unknown-but-valid addresses still return 200** (not 404) — no CLI exit-code regression.
- Domain regex mirrors upstream exactly; accepts SNS `.sol`, ANS custom TLDs, underscores (`my_wallet.sol`), and multi-label subdomains (`sub.toly.sol`); rejects length > 255.

## Test plan

- [ ] `helius wallet identity 86xCnPeV69n6t3DnyGvkKobf9FdN2H9oiVDdaMpo2MMY` (address — unchanged behavior)
- [ ] `helius wallet identity toly.sol` (SNS resolves on mainnet)
- [ ] `helius wallet identity helius.bonk` (ANS resolves on mainnet)
- [ ] `helius wallet identity bogus.sol` → exit 55 (NOT_FOUND)
- [ ] `helius wallet identity not-a-real-thing` → exit 52 (INVALID_ADDRESS, client-side)
- [ ] `helius wallet identity toly.sol --network devnet` → exit 53 (INVALID_INPUT, "Domain resolution is only available on mainnet")
- [ ] `helius wallet identity-batch toly.sol 86xC...MMY --json` → mixed JSON output with `inputDomain` on the domain entry
- [ ] MCP inspector: `getWalletIdentity { address: "toly.sol" }` returns identity; tool description advertises SNS/ANS
- [ ] CI sync-check passes (canonical vs `helius-plugin/`/`helius-cursor/` mirrors all clean)
- [ ] `pnpm -C helius-cli build && pnpm -C helius-mcp build` clean

### DIFF
--- a/helius-cli/src/commands/wallet.ts
+++ b/helius-cli/src/commands/wallet.ts
@@ -2,14 +2,14 @@ import chalk from "chalk";
 import { resolveApiKey, restRequest, type ResolveOptions } from "../lib/helius.js";
 import { formatAddress, formatTable, formatEnumLabel, type TableColumn } from "../lib/formatters.js";
 import { outputJson, exitWithError, handleCommandError, createSpinner, withRetry, type OutputOptions, type RetryOptions } from "../lib/output.js";
-import { validateAddress, validateAddresses } from "../lib/validation.js";
+import { validateAddress, validateAddressOrDomain, validateAddressesOrDomains } from "../lib/validation.js";
 
 interface WalletOptions extends OutputOptions, ResolveOptions, RetryOptions {}
 
 export async function walletIdentityCommand(address: string, options: WalletOptions = {}): Promise<void> {
   const spinner = createSpinner(options);
   try {
-    const addrErr = validateAddress(address);
+    const addrErr = validateAddressOrDomain(address);
     if (addrErr) exitWithError("INVALID_ADDRESS", addrErr, undefined, !!options.json);
     spinner?.start("Resolving API key...");
     const apiKey = await resolveApiKey(options);
@@ -34,7 +34,7 @@ export async function walletIdentityCommand(address: string, options: WalletOpti
 export async function walletIdentityBatchCommand(addresses: string[], options: WalletOptions = {}): Promise<void> {
   const spinner = createSpinner(options);
   try {
-    const addrErr = validateAddresses(addresses);
+    const addrErr = validateAddressesOrDomains(addresses);
     if (addrErr) exitWithError("INVALID_ADDRESS", addrErr, undefined, !!options.json);
     spinner?.start("Resolving API key...");
     const apiKey = await resolveApiKey(options);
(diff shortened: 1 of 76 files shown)


helius-mcp/src/tools/wallet.ts
```diff
@@ -64,23 +64,30 @@ export function registerWalletTools(server: McpServer) {
 
         const lines = [`**Batch Identity Lookup** (${results.length} results)`, ''];
 
-        for (const entry of results) {
+        for (const entry of results as Array<any>) {
```
0xIchigo, line 70: We should leave a TODO comment here to change from `any` once we add the relevant types to the SDK

helius-mcp/src/tools/wallet.ts
```diff
@@ -40,9 +40,9 @@ export function registerWalletTools(server: McpServer) {
   // ─── Batch Wallet Identity ───
   server.tool(
     'batchWalletIdentity',
-    'BEST FOR: identifying multiple wallets at once (up to 100). PREFER getWalletIdentity for a single address. Look up identities for up to 100 Solana addresses. Returns names, types, and categories. Credit cost: 100 credits.',
+    'BEST FOR: identifying multiple wallets at once (up to 100). PREFER getWalletIdentity for a single address. Look up identities for up to 100 entries. Returns names, types, and categories. Entries may be addresses or SNS/ANS domains (mainnet only); unresolved domains are returned with `unresolved: true`. Credit cost: 100 credits.',
     {
-      addresses: z.array(z.string()).describe('Array of Solana wallet addresses (base58 encoded, max 100)')
+      addresses: z.array(z.string()).describe('Array of up to 100 entries — each a base58 address or SNS/ANS domain')
```
0xIchigo: Nit: the name here is stale, as we now refer to them as "entries" instead of "addresses"

helius-cli/src/commands/wallet.ts
```diff
@@ -48,11 +48,22 @@ export async function walletIdentityBatchCommand(addresses: string[], options: W
     const identities = Array.isArray(result) ? result : [];
     console.log(chalk.bold(`\nWallet Identities (${identities.length}):\n`));
     for (const id of identities) {
-      if (id.name || id.type) {
-        console.log(`  ${chalk.cyan(id.address || "?")} - ${chalk.green(id.name || "Unknown")} (${id.type ? formatEnumLabel(id.type) : "N/A"})`);
+      if (id.unresolved) continue;
+      const subject = id.inputDomain && id.address
+        ? `${chalk.cyan(id.inputDomain)} ${chalk.gray("→")} ${chalk.cyan(id.address)}`
+        : chalk.cyan(id.address || id.inputDomain || "?");
+      if (id.name || (id.type && id.type !== "unknown")) {
```
0xIchigo: A domain that resolves cleanly to a wallet with no Orb identity falls through to the "N wallet(s) have no known identity" counter. This means the user never sees which domain resolved to which public key

We should still print rows when `id.inputDomain` is set, even without a name or type

0xIchigo (changes requested):
Based, just a few comments. Also:
- When the input is a domain, the CLI echoes the domain and never surfaces the resolved pubkey (which is present on result.address). We should mirror the batch path's inputDomain -> address rendering here so the user sees the resolution. The PR description only calls out batch output updates, and the single-endpoint formatter was likely missed
- The MCP also has a similar issue for domain inputs, as it renders **Address:** toly.sol instead of the resolved pubkey. LLMs may then feed the domain string into downstream tools that expect base58. We should use `data.address` (falling back to the input) and optionally add an **Input:** toly.sol line when data.inputDomain is set

0xIchigo (approved):
LGTM 🚀

PR #107, opened 2026-04-23
feat(cli): Always emit suggestion in JSON error envelopes

nathanliow:
## Summary

- Populate `CLI_GUIDANCE` for the 5 error codes without entries (`AUTH_FAILED`, `PROJECT_EXISTS`, `API_ERROR`, `SDK_ERROR`, `INVALID_ADDRESS`). `SDK_ERROR` is the common unclassified-error bucket, so this was a visible gap.
- Make `ErrorEnvelope.suggestion` required (drop the `?`) and emit it unconditionally in `buildErrorEnvelope()`, with a generic "re-run with --debug" fallback for any code that ever slips past the map.
- Docs: `README.md` Output Format section and `CLAUDE.md` envelope-shape comments now say `suggestion` is always present.

### Why

Agents get `{ error, recoverable, error_code, category }` on `--json` failure and today have to parse the human `error` string to know what to do. With `suggestion` guaranteed to be a short actionable next step (command to run, config to fix, etc.), agents can self-correct or surface the hint directly.

### Not changed

- `README.md:11` migration note still shows the v1.3 `suggestion?` shape for historical accuracy.
- `ws.ts:52` `handleWsError()` uses an ad-hoc object literal (not `ErrorEnvelope`), so the tighter type won't catch it. Existing conditional spread stays — worth aligning in a follow-up.

## Test plan

- [x] `tsc --noEmit` clean on changed code (pre-existing `signup.ts:217` error is on `main`, unrelated).
- [x] `helius balance --json not-a-real-address` → `suggestion` present on `INVALID_ADDRESS` (was missing before).
- [x] `HELIUS_API_KEY=bad helius balance --json <pubkey>` → `suggestion` still present on `INVALID_API_KEY`.
- [x] `helius tx parse --json BADSIG` → `suggestion` present on `INVALID_INPUT`.
- [x] `rg '"ok":\s*false|ok:\s*false' src/` confirms `buildErrorEnvelope()` is the only error-envelope emitter.

### DIFF
--- a/helius-cli/CLAUDE.md
+++ b/helius-cli/CLAUDE.md
@@ -47,7 +47,7 @@ All data commands follow this pattern:
 3. Get SDK client via `getClient(apiKey, network)`
 4. Call SDK method
 5. Output: formatted terminal (chalk) or `outputJson(payload)` for `--json`. As of v1.3, `outputJson()` auto-wraps its argument in `{ ok: true, v: 1, data: <payload> }` — callers still pass the raw payload.
-6. Error: use `handleCommandError(error, options, spinner)` — classifies the error, emits the error envelope (`{ ok: false, v: 1, error_code, category, error, recoverable, suggestion?, details? }`) or spinner output, and exits with the classified exit code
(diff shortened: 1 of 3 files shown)


0xIchigo (approved):
LGTM :shipit:

PR #108, opened 2026-04-28
fix mcp config recovery, telemetry, and laserstream api handling

mooncitydev:
hey, went through the mcp and cli telemetry paths and a few things stood out.

first, getLaserstreamUrl was calling getApiKey() even though the returned URL does not embed the key (the laserstream sdk takes apiKey in subscribe options). that meant getLaserstreamInfo could throw NO_API_KEY before showing public endpoint text. i dropped the unused getApiKey call and gate laserstreamSubscribe with the same noApiKeyResponse() pattern as the other rpc tools.

second, posthog delivery used .catch(() => { feedbackEnabled = false }). any single fetch failure flipped the flag off for the whole process so every later analytics event became a noop. flaky network or relay hiccup should not permanently silence telemetry.

third, helius-cli already repairs corrupted ~/.helius/config.json with regex salvage. helius-mcp loadConfig silently returned {} on bad json, so api keys nested in corrupt files disappeared without clues. mirrored the salvage + stderr warnings + repaired save behavior for aligned config between cli and mcp.

cheers
moon

### DIFF
--- a/helius-cli/src/lib/feedback.ts
+++ b/helius-cli/src/lib/feedback.ts
@@ -10,7 +10,6 @@ const HELIUS_DIR = path.join(os.homedir(), '.helius');
 const KEYPAIR_PATH = path.join(HELIUS_DIR, 'keypair.json');
 const WALLET_CACHE_PATH = path.join(HELIUS_DIR, 'wallet-address');
 
-let feedbackEnabled = true;
 let walletAddress: string | null = null;
 let identifySent = false;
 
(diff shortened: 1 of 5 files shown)


0xIchigo (commented):
Thank you for your contribution @mooncitydev ! This looks good—I just have two nits:
- The `network` regex `/"network"\s*:\s*"([^"]+)"/` is wider than the CLI's (`mainnet|devnet`). Since the MCP only accepts mainnet-beta/devnet URLs at runtime, tightening to (`mainnet|mainnet-beta|devnet`) would prevent "recovering" an unusable value
- The try / catch block around `getLaserstreamUrl()` in `getLaserstreamInfo` is now unreachable, as the function no longer throws. This could be simplified with a direct call

0xIchigo (approved):
LGTM :shipit:

mooncitydev:
would love to connect with you on discord
mine is mooncityorg, you can find me helius server 
tyvm for merging

PR #109, opened 2026-05-02
feat(cli+mcp): unified crypto checkout — signup / upgrade / credits / pay

nathanliow:
## Summary

Phases 1, 2, and 3 of the SDK signup revamp on the CLI + MCP side. The CLI gains link-mode UX for `signup` / `upgrade` / `pay`, plus a new `helius credits` command. The MCP migrates fully off the legacy `agenticSignup` tool to the new `signup` (with `mode: "link" | "autopay" | "resume"`) tool, and gains parallel `upgradePlan` / `purchaseCredits` / `payRenewal` tools.

Pairs with helius-labs/helius-sdk#312 and helius-labs/monorepo#15009. Don't merge until both upstream PRs are in prod.

## helius-cli — every paid command now has three modes

| Command | Default (link) | `--pay` (autopay) | `--resume` |
|---|---|---|---|
| `helius signup` | print payment URL, save pendingSignup | send USDC + memo, poll, provision | poll stored pending intent |
| `helius upgrade` | print upgrade payment URL, save pendingUpgrade | autopay + poll | poll stored pending upgrade |
| `helius pay <id>` | print renewal payment URL | autopay + poll (no activation gate) | poll the renewal intent |
| `helius credits` (new) | print credits-topup URL, save pendingCredits | autopay + poll | poll stored pending credits |

JSON shapes are uniform across commands: `PAYMENT_REQUIRED` / `SUCCESS` / `PENDING` / `EXPIRED` / `FAILED` / `MISSING_PENDING_*` / `ALREADY_SUBSCRIBED` / `UPGRADE_REQUIRED`. Pending state in `~/.helius/config.json` is per-flow (`pendingSignup`, `pendingUpgrade`, `pendingCredits`) so the three flows don't collide. Local `expiresAt` is informational; cleanup only on backend-reported terminal states or `--restart`.

## helius-mcp — agenticSignup tool retired, full Phase 3 migration

- **`signup` tool** replaces `agenticSignup`. Takes `mode: "link" | "autopay" | "resume"` (default `"link"`). Link mode prints the payment URL, autopay sends USDC + memo from the local keypair, resume polls a pending intent.
- **`upgradePlan`, `purchaseCredits`, `payRenewal`** tools rewritten on the new SDK helpers. Each takes `mode: "link" | "autopay"`.
- **`getStarted`** rewritten around the link flow (drops "fund wallet → check balance" steps).
- **`checkSignupBalance`** tool removed — no longer needed for browser-pay flow.
- Router catalog + action lists updated.

## Other changes

- `lib/api.ts` exposes pass-throughs for the new SDK Phase 2 helpers (`upgradePlan`/`AndPay`, `purchaseCredits`/`AndPay`, `payRenewal`/`AndPay`).
- `lib/config.ts` factored into `PendingPaymentBase` shared shape + flow-specific extensions (`PendingSignup` / `PendingUpgrade` / `PendingCredits`).
- `bin/helius.ts` flag changes for the new modes; new `helius credits` command registration.
- README updated for Phase 1 link mode (Phase 2 docs follow once SDK ships).

## Coordination

- Pairs with helius-labs/helius-sdk#312 and helius-labs/monorepo#15009 and helius-labs/dev-portal-v2#1088.
- After SDK publishes to npm, bump `helius-sdk` dep version on this branch before merging.

## Test plan

- [ ] CLI Phase 1: `pnpm test` — 17 / 17 (pre-existing test for signup flow still green)
- [ ] MCP: `vitest run` — 223 / 223 (existing suite still passes after the auth.ts rewrite)
- [ ] CLI typecheck: `tsc --noEmit` clean (one pre-existing `reclaim.ts` error unrelated)
- [ ] MCP typecheck: clean
- [ ] Phase 1 smoke verified live against dev-api-dev (signup link → browser pay → resume)
- [ ] Phase 2 / Phase 3 live smoke pending the SDK npm publish

### DIFF
--- a/helius-cli/src/commands/credits.test.ts
+++ b/helius-cli/src/commands/credits.test.ts
@@ -3,6 +3,7 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 const mockPurchaseCredits = vi.fn();
 const mockPayPaymentLink = vi.fn();
 const mockGetPaymentStatus = vi.fn();
+const mockGetPaymentIntent = vi.fn();
 const mockListProjects = vi.fn();
 const mockWalletSignup = vi.fn();
 
@@ -10,6 +11,7 @@ vi.mock("../lib/api.js", () => ({
   walletSignup: (...args: unknown[]) => mockWalletSignup(...args),
   listProjects: (...args: unknown[]) => mockListProjects(...args),
   getPaymentStatus: (...args: unknown[]) => mockGetPaymentStatus(...args),
+  getPaymentIntent: (...args: unknown[]) => mockGetPaymentIntent(...args),
   payPaymentLink: (...args: unknown[]) => mockPayPaymentLink(...args),
   purchaseCredits: (...args: unknown[]) => mockPurchaseCredits(...args),
 }));
@@ -187,6 +189,23 @@ describe("credits command", () => {
     const { json } = await captureStdout(() => creditsCommand({ ...baseOpts, resume: true } as any));
     expect(json?.data.status).toBe("SUCCESS");
     expect(json?.data.qty).toBe(1);
+    expect(mockClearPendingCredits).toHaveBeenCalled();
+  });
+
+  it("--resume SUCCESS without stored txSignature backfills it from getPaymentIntent", async () => {
+    mockGetPendingCredits.mockReturnValue({ ...samplePending });
+    mockGetPaymentStatus.mockResolvedValue({
+      status: "completed",
+      phase: "complete",
+      subscriptionActive: true,
+      readyToRedirect: true,
+      message: "ok",
+    });
+    mockGetPaymentIntent.mockResolvedValue({ txSignature: "tx-from-intent" });
+    const { json } = await captureStdout(() => creditsCommand({ ...baseOpts, resume: true } as any));
+    expect(json?.data.status).toBe("SUCCESS");
+    expect(json?.data.txSignature).toBe("tx-from-intent");
+    expect(mockGetPaymentIntent).toHaveBeenCalledWith(samplePending.jwt, samplePending.paymentIntentId);
     expect(mockClearPendingCredits).toHaveBeenCalled();
   });
 
--- a/helius-cli/src/commands/credits.test.ts
+++ b/helius-cli/src/commands/credits.test.ts
@@ -1,0 +1,206 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const mockPurchaseCredits = vi.fn();
+const mockPayPaymentLink = vi.fn();
+const mockGetPaymentStatus = vi.fn();
+const mockListProjects = vi.fn();
+const mockWalletSignup = vi.fn();
+
+vi.mock("../lib/api.js", () => ({
+  walletSignup: (...args: unknown[]) => mockWalletSignup(...args),
+  listProjects: (...args: unknown[]) => mockListProjects(...args),
+  getPaymentStatus: (...args: unknown[]) => mockGetPaymentStatus(...args),
+  payPaymentLink: (...args: unknown[]) => mockPayPaymentLink(...args),
+  purchaseCredits: (...args: unknown[]) => mockPurchaseCredits(...args),
+}));
+
+const mockSetJwt = vi.fn();
+const mockGetPendingCredits = vi.fn();
+const mockSetPendingCredits = vi.fn();
+const mockUpdatePendingCredits = vi.fn();
+const mockClearPendingCredits = vi.fn();
+
+vi.mock("../lib/config.js", () => ({
+  setJwt: (...args: unknown[]) => mockSetJwt(...args),
+  getPendingCredits: () => mockGetPendingCredits(),
+  setPendingCredits: (...args: unknown[]) => mockSetPendingCredits(...args),
+  updatePendingCredits: (...args: unknown[]) => mockUpdatePendingCredits(...args),
+  clearPendingCredits: () => mockClearPendingCredits(),
+  SHARED_CONFIG_PATH: "/tmp/test-config.json",
+}));
+
+const mockKeypairExists = vi.fn();
+vi.mock("./keygen.js", () => ({
+  keypairExists: (...args: unknown[]) => mockKeypairExists(...args),
+  getDefaultKeypairPath: () => "/tmp/keypair.json",
+}));
+
+const mockLoadKeypair = vi.fn();
+const mockGetAddress = vi.fn();
+const mockSignAuthMessage = vi.fn();
+vi.mock("../lib/wallet.js", () => ({
+  loadKeypairFromFile: (...args: unknown[]) => mockLoadKeypair(...args),
+  getAddress: (...args: unknown[]) => mockGetAddress(...args),
+  signAuthMessage: (...args: unknown[]) => mockSignAuthMessage(...args),
+}));
+
+const mockCheckSolBalance = vi.fn();
+const mockCheckUsdcBalance = vi.fn();
+vi.mock("../lib/payment.js", () => ({
+  checkSolBalance: (...args: unknown[]) => mockCheckSolBalance(...args),
+  checkUsdcBalance: (...args: unknown[]) => mockCheckUsdcBalance(...args),
+}));
+
+vi.mock("../lib/feedback.js", () => ({
+  sendCommandEvent: vi.fn(),
+  getCurrentCommand: () => "credits",
+}));
+
+import { creditsCommand } from "./credits.js";
+
+interface CapturedJson {
+  ok: boolean;
+  v?: number;
+  data?: any;
+  error?: any;
+}
+
+function captureStdout(fn: () => Promise<void>) {
+  const chunks: string[] = [];
+  let exitCode: number | null = null;
+
+  const logSpy = vi.spyOn(console, "log").mockImplementation((...args: unknown[]) => {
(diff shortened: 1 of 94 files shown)


helius-mcp/src/tools/config.ts
```diff
@@ -11,7 +11,7 @@ export function registerConfigTools(server: McpServer) {
     'setHeliusApiKey',
     keyFromEnv
       ? 'API key is already configured via environment. You do NOT need to call this tool - just use the other Helius tools directly.'
-      : 'Set an existing Helius API key for the current session. If the user does not have a key, use the agentic signup flow instead: generateKeypair → fund wallet → agenticSignup. Get a key at https://dashboard.helius.dev/api-keys',
+      : 'Set an existing Helius API key for the current session. If the user does not have a key, use the signup flow instead: generateKeypair → fund wallet → signup. Get a key at https://dashboard.helius.dev/api-keys',
```
0xIchigo: Nit: we're still using some legacy phrasing here. We should drop "fund wallet". Link-mode signup doesn't require pre-funding the keypair. We could say something like "use the signup flow: generateKeypair → signup (link mode prints a payment URL)"

helius-mcp/src/tools/plans.ts
```diff
@@ -191,7 +205,7 @@ export function registerPlanTools(server: McpServer) {
           `**Auth:** Not authenticated\n\n` +
           `No API key or session found. To get started:\n` +
           `- If you have a key: use the \`setHeliusApiKey\` tool\n` +
-          `- If you need an account: use \`generateKeypair\` → fund wallet → \`agenticSignup\``
+          `- If you need an account: use \`generateKeypair\` → fund wallet → \`signup\``
```
0xIchigo: Same legacy phrasing as `config.ts:14`. We should drop "fund wallet" from the not-authenticated branch

helius-cli/src/commands/signup.ts
```diff
@@ -201,119 +197,494 @@ export async function signupCommand(options: SignupOptions): Promise<void> {
...
+    return;
+  }
+  if (status.status === "completed" && !status.readyToRedirect) {
+    // Payment already settled; just keep polling.
+    await pollAndProvision(stored, undefined, options, spinner);
+    return;
+  }
+
+  // 3. status === "pending" → send USDC.
+  const secretKey = freshSecretKey ?? (await loadSecretKey(options));
```
0xIchigo: If the user runs helius signup (link) on Monday, rotates `id.json`, then runs `helius signup --pay` Tuesday, this loads the new keypair and sends USDC from a wallet that doesn't match `stored.walletAddress`, and  memo binding fails on the backend. Maybe we should assert `(await getAddress(keypair)) === stored.walletAddress` before `payPaymentLink`, with an error pointing at `--restart` (i.e., the same pattern in `upgrade.ts:238` and `credits.ts:210`)

helius-cli/src/commands/credits.test.ts
```diff
@@ -0,0 +1,206 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const mockPurchaseCredits = vi.fn();
+const mockPayPaymentLink = vi.fn();
+const mockGetPaymentStatus = vi.fn();
+const mockListProjects = vi.fn();
+const mockWalletSignup = vi.fn();
+
+vi.mock("../lib/api.js", () => ({
```
0xIchigo, line 10: The mock doesn't expose `getPaymentIntent`, so the `txSignature` backfill added is not a function and is silently swallowed by the `try {}` in `emitSuccess`, so the backfill is effectively untested. We should add `getPaymentIntent: vi.fn()` and one assertion that a `--resume SUCCESS` without a stored `txSignature` resolves it from the intent

helius-cli/src/commands/pay.ts
```diff
@@ -1,108 +1,303 @@
...
+        options,
+        spinner,
       );
+      return;
     }
-    spinner?.succeed("Payment complete");
+
+    // Default: print the renewal link.
+    spinner?.start("Fetching renewal payment intent...");
+    const result = await sdkPayRenewal(auth.token, paymentIntentId);
```
0xIchigo, line 95: We should switch on `result.kind` here. If `payRenewal` ever returns a `non-payment_required` kind, this crashes

helius-cli/src/commands/upgrade.ts
```diff
@@ -1,189 +1,483 @@
...
       const emailErr = validateEmail(options.email);
       if (emailErr) exitWithError("INVALID_INPUT", emailErr, undefined, !!options.json);
     }
 
-    // Check keypair exists
-    if (!keypairExists(options.keypair)) {
-      exitWithError("KEYPAIR_NOT_FOUND", `Keypair not found at ${options.keypair}`, undefined, !!options.json);
+    if (options.restart) clearPendingUpgrade();
+
+    const stored = getPendingUpgrade();
```
0xIchigo, line 83: `signup.ts` has `refreshStoredFromBackend` that handles "local expiresAt is past — query backend before reprinting", with three outcomes: `cleared` / `provisioned` / `still_payable`. `upgrade` and `credits` skip this entirely and just reprint `emitPaymentRequired(stored, true)`, so if the backend has expired/failed/completed the upgrade intent, the user keeps getting served a stale URL on every re-run until they remember `--restart`. Either lift `refreshStoredFromBackend` into a shared helper used by all three flows, or document that upgrade/credits intentionally trust local expiry

helius-mcp/src/tools/auth.ts
```diff
@@ -205,208 +170,219 @@ export function registerAuthTools(server: McpServer) {
...
           discoveryPath,
           frictionPoints,
         });
       }
+
       try {
+        // Resume mode: just poll the existing JWT/intent. Caller must have
+        // already gone through link mode in a previous invocation; the JWT
+        // from that invocation is on disk.
+        if (mode === 'resume') {
```
0xIchigo, line 222: "resume" only checks `getJwt()`, meaning that it doesn't have a `pendingSignup` equivalent to the CLI's, so it can't distinguish "pending signup awaiting browser pay" from "stale JWT left over from a completed signup ages ago". The fallback ("Payment not yet provisioned … try again in a moment") is the same for both. It's probably fine for now, since the MCP is more stateless, but worth noting in the resume tool description that it strictly polls the JWT-bound project list. A user/agent who calls resume after a long-completed signup will get confusing output

0xIchigo (changes requested):

helius-mcp/src/tools/config.ts
nathanliow: Updated: now reads `generateKeypair → signup (link mode prints a payment URL)`. 99d00d0

helius-mcp/src/tools/plans.ts
nathanliow: Updated to `generateKeypair → signup (link mode prints a payment URL)`. 99d00d0

helius-cli/src/commands/signup.ts
nathanliow: Added the wallet-mismatch guard in all three flows. `runPayWithStored` now derives `payerAddress = await getAddress(keypair)` and asserts it equals `stored.walletAddress` (signup) or `stored.payerWallet` (upgrade/credits — added to `PendingUpgrade`/`PendingCredits`, populated at intent creation). On mismatch we exit `INVALID_INPUT` pointing at `--restart`. For backwards compatibility with intents stored before this change, `payerWallet` is optional on upgrade/credits and the check is skipped when undefined (one-version migration window). 99d00d0

helius-cli/src/commands/credits.test.ts
nathanliow: Added `getPaymentIntent: vi.fn()` to the `../lib/api.js` mock and a new test `--resume SUCCESS without stored txSignature backfills it from getPaymentIntent` that asserts `json.data.txSignature === "tx-from-intent"` and that `getPaymentIntent` was called with the stored jwt + paymentIntentId. 99d00d0

helius-cli/src/commands/pay.ts
nathanliow: Done at both `sdkPayRenewal` call sites (link-mode default and `runPay`). Used the exhaustive `switch` + `const _exhaustive: never = result.kind` pattern so it compiles cleanly today (`PayRenewalResult` is a single-variant union) and stays correct if the SDK adds variants. 99d00d0

helius-cli/src/commands/upgrade.ts
nathanliow: Lifted: extracted `checkBackendForRefresh(jwt, paymentIntentId, spinner)` into `helius-cli/src/lib/payment.ts` returning `{ verdict: "cleared" | "completed" | "still_payable" }`. `upgrade` and `credits` now call it on local-expired stored intents and act on the verdict (`completed` → poll-and-emit, `cleared` → clear + fall through to fresh-intent path, `still_payable` → reprint). Signup's existing `refreshStoredFromBackend` now delegates to the helper and translates `completed` → `provisioned` (poll-and-provision before returning). 99d00d0

helius-mcp/src/tools/auth.ts
nathanliow: Updated the `signup` tool description in `helius-mcp/src/tools/auth.ts` to note: "resume strictly polls the JWT-bound project list and cannot distinguish 'pending signup awaiting browser pay' from 'stale JWT left over from a long-completed signup' — if the wallet already has an API key, prefer `setHeliusApiKey` over re-resuming." 99d00d0

0xIchigo (approved):
LGTM :shipit:

PR #114, opened 2026-05-12
docs(plugin): Advertise Jupiter/OKX/SVM Skills and Bump to 1.1

0xIchigo:
This PR updates `helius-plugin/README.md` to list the Jupiter, OKX, and SVM skills under "What's included" (previously only build/DFlow/Phantom were advertised, even though all three skills already ship in `helius-plugin/skills/`). We also bump `helius-plugin/.claude-plugin/plugin.json` from `1.0.0` to `1.1.0` to reflect the expanded surface


nathanliow (approved):
LGTM

PR #116, opened 2026-05-18
fix(cli): tx history --before now uses beforeSignature

mooncitydev:
hey team, noticed the cli path for `helius tx history` was passing `before` into `enhanced.getTransactionsByAddress` but the sdk only reads `beforeSignature`. extra keys just get ignored so pagination never moved and you would keep seeing the first page.

mapped the flag to `beforeSignature` and typed the params as `GetEnhancedTransactionsByAddressRequest` so the shape stays honest.

## test plan
- `npm run build` in `helius-cli`
- optional: `helius tx history <addr> --before <sig>` and confirm results change vs the first page

### DIFF
--- a/helius-cli/src/commands/tx.ts
+++ b/helius-cli/src/commands/tx.ts
@@ -1,4 +1,5 @@
 import chalk from "chalk";
+import type { GetEnhancedTransactionsByAddressRequest } from "helius-sdk/enhanced/types";
 import { setupClient, type ResolveOptions } from "../lib/helius.js";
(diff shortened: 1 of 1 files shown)


0xIchigo (approved):
Thank you for your contribution! 🔥

0xIchigo:
@mooncitydev can you please [sign your commit](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) so I can merge this into the repo?

PR #118, opened 2026-05-29
feat(helius-cli): OAuth/PKCE login + logout + whoami (v2.0.0)

nathanliow:
## Summary
Browser-based `helius login` (OAuth/PKCE) for users with existing dashboard accounts. **Supports Google, GitHub, email/password, and wallet sign-in.** Symmetric `logout` + `whoami` round out the auth surface.

## Cross-repo context (merge in this order)
1. Backend: https://github.com/helius-labs/monorepo/pull/15655
2. SDK: https://github.com/helius-labs/helius-sdk/pull/329
3. Dev portal: https://github.com/helius-labs/dev-portal-v2/pull/1161
4. **This PR** (CLI)

## ⚠️ Before merging / publishing v2.0.0
1. Merge & publish the helius-sdk PR (new minor — exposes `oauthTokenExchange`, `HELIUS_API_URL` env, JSON error parsing).
2. Bump `helius-sdk` dep in `helius-cli/package.json` from `^2.2.2` to the new version.
3. Verify `pnpm build` succeeds. Pre-existing errors in `signup.ts` / `checkout.ts` reference SDK functions removed in an unrelated migration — **out of scope here**, separate ticket.
4. Publish `helius-cli@2.0.0`.

## Test plan
- [ ] 6/6 vitest tests pass (`pnpm test`)
- [ ] End-to-end with Google sign-in
- [ ] End-to-end with wallet sign-in (full wallet pubkey on hover, shortened in main text)
- [ ] `helius logout` clears jwt, keeps apiKey; `--all` wipes everything
- [ ] `helius whoami --verify` confirms backend
- [ ] `helius login` after `logout` re-authenticates cleanly
- [ ] Already-logged-in short-circuit prints `Already logged in as ...` and exits 0

## What's outside scope (documented, not fixed)
- The free-plan project-create gate still rejects `authSource='cli'`. CLI users currently create projects via the dashboard, then use the CLI for management. Separate ticket if we want CLI-from-cli project creation.
- WorkOS SSO users rejected at `/oauth/token` with a clear error — domain lockdown was already excluding them. Same posture as before.
- Pre-existing `signup.ts` / `checkout.ts` build errors against the local SDK v3.0.0 — unrelated stale references.

### DIFF
--- a/helius-cli/CHANGELOG.md
+++ b/helius-cli/CHANGELOG.md
@@ -1,17 +1,5 @@
 # Changelog
 
-> **⚠️ Before merging v2.0.0 / publishing to npm:** this release depends on
-> `oauthTokenExchange`, `HELIUS_API_URL` env override, and JSON error parsing
-> being available in `helius-sdk`. Those land in a separate SDK PR. Sequence:
->
-> 1. Merge & publish the `helius-sdk` PR (new minor/major).
-> 2. Bump `helius-sdk` dep in `package.json` from `^2.2.2` to the new version.
-> 3. Verify `pnpm build` succeeds, then publish `helius-cli` v2.0.0.
->
-> Pre-existing build errors in `src/commands/signup.ts` and `src/lib/checkout.ts`
-> reference SDK functions (`agenticSignup`, `executeCheckout`, etc.) that were
-> removed in a separate, unrelated SDK migration — out of scope for this PR.
-
 ## 2.0.0 — Breaking change: OAuth/PKCE login
 
 `helius login` now opens your browser to authenticate via dashboard.helius.dev, instead of signing a message with a local Solana keypair.
--- a/helius-cli/CHANGELOG.md
+++ b/helius-cli/CHANGELOG.md
@@ -1,0 +1,42 @@
+# Changelog
+
+> **⚠️ Before merging v2.0.0 / publishing to npm:** this release depends on
+> `oauthTokenExchange`, `HELIUS_API_URL` env override, and JSON error parsing
+> being available in `helius-sdk`. Those land in a separate SDK PR. Sequence:
+>
+> 1. Merge & publish the `helius-sdk` PR (new minor/major).
+> 2. Bump `helius-sdk` dep in `package.json` from `^2.2.2` to the new version.
+> 3. Verify `pnpm build` succeeds, then publish `helius-cli` v2.0.0.
+>
+> Pre-existing build errors in `src/commands/signup.ts` and `src/lib/checkout.ts`
+> reference SDK functions (`agenticSignup`, `executeCheckout`, etc.) that were
+> removed in a separate, unrelated SDK migration — out of scope for this PR.
+
+## 2.0.0 — Breaking change: OAuth/PKCE login
+
+`helius login` now opens your browser to authenticate via dashboard.helius.dev, instead of signing a message with a local Solana keypair.
+
+### Breaking
+- **`helius login -k <keypair>` / `--keypair` flag removed.** The new flow uses OAuth 2.0 Authorization Code Grant + PKCE (RFC 7636 + RFC 8252) — same pattern as `aws sso login`, `gh auth login`, `vercel login`. Wallet-based account creation continues to work via `helius signup`; only the auth-token retrieval path changed.
+- If you previously authenticated with `helius login -k`, your cached JWT in `~/.helius/config.json` keeps working until it expires. The next time you need to re-authenticate, run `helius login` (no flag).
+
+### New
+- `helius login` — browser-based authentication for users with existing dashboard.helius.dev accounts. Supports four auth methods: email/password, Google, GitHub, and **wallet** (Solana signature).
+- `helius login --no-browser` — prints the URL for manual paste (useful in remote/headless environments).
+- `helius login --json` — emits `{ status: "AWAITING_CALLBACK", verification_uri }` while waiting and `{ status: "SUCCESS", user, expires_at }` on completion.
+- Already-logged-in short-circuit: a second `helius login` with a valid JWT prints `Already logged in as <email>.` (or `wallet <pubkey>` for wallet sessions) and exits 0.
+- **`helius logout`** — clears the local session token. Default keeps the API key so RPC calls still work; `--all` wipes the entire config.
+- **`helius whoami`** — shows the authenticated identity (email or wallet pubkey), session expiry, project, API key, and network. `--verify` pings the backend to confirm the token is still valid server-side. `--json` for scripting.
+- "No projects found" guidance updated across `projects`, `status`, `apikeys` to point users at the dashboard for project creation (the CLI cannot create projects for OAuth-authenticated users — that gate is enforced backend-side and is a separate ticket).
+
+### Not supported
+- **WorkOS SSO users**: rejected at the backend with a clean error message. Enterprise SSO users typically have their CLI access provisioned via the org admin portal; this is consistent with the existing domain-lockdown check that already rejects SSO at `/oauth/token`.
+
+### Security
+- PKCE binds each auth code to the originating CLI invocation. Public OAuth client per RFC 8252 §6 — no client secret.
+- Loopback redirect on `127.0.0.1` (literal IPv4, not `localhost`) per RFC 8252 §7.3, random ephemeral port via `:0`.
+- 5-minute timeout on the loopback listener; CSRF defense via `state` round-trip.
+
+## 1.3.0
+
+Previous releases — see git history.
--- a/helius-cli/src/commands/login.ts
+++ b/helius-cli/src/commands/login.ts
@@ -1,50 +1,123 @@
 import chalk from "chalk";
-import { loadKeypairFromFile, signAuthMessage, getAddress } from "../lib/wallet.js";
-import { signup } from "../lib/api.js";
-import { setJwt } from "../lib/config.js";
-import { keypairExists } from "./keygen.js";
-import { outputJson, exitWithError, handleCommandError, createSpinner, type OutputOptions } from "../lib/output.js";
+import { generatePkcePair, generateState, describeJwtIdentity } from "../lib/oauth.js";
+import { startLoopback } from "../lib/loopback-server.js";
+import { openInBrowser } from "../lib/browser.js";
+import { oauthTokenExchange, listProjects } from "../lib/api.js";
+import { getJwt, setJwt } from "../lib/config.js";
+import {
+  outputJson,
+  handleCommandError,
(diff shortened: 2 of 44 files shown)


helius-cli/src/lib/loopback-server.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+    if (!code) {
+      res.writeHead(400, { "Content-Type": "text/html" });
+      res.end(ERROR_PAGE("missing authorization code"));
+      rejectCb(new Error("NO_CODE"));
+      return;
+    }
+
+    res.writeHead(200, { "Content-Type": "text/html" });
+    res.end(SUCCESS_PAGE);
```
0xIchigo, line 81: The handler sends `SUCCESS_PAGE` and resolves the promise immediately, meaning state comparison only happens later in `awaitCallback`. So, for example, on a state mismatch, the browser could still show "Logged in" while the CLI aborts with `STATE_MISMATCH`. We should be validating `state` inside the request handler (i.e., serve the error page on mismatch, and keep listening or 400 rather than resolving) so the browser reflects the real outcome, and a bad early hit doesn't poison the flow

helius-cli/src/lib/loopback-server.ts
```diff
@@ -0,0 +1,90 @@
...
+    resolveCb(`${code}|${state ?? ""}`);
```
0xIchigo: `${code}|${state}` and `result.split("|")` assumes that neither value contains a pipe. `state` is base64url, so it's safe, but the OAuth authorization `code` is an opaque, server-defined string with no guaranteed charset. We could just resolve a structured value (`{ code, state }`) instead of joining and splitting

helius-cli/src/lib/loopback-server.ts
```diff
@@ -0,0 +1,90 @@
...
+      try {
+        const result = await Promise.race([cbPromise, timeoutPromise]);
+        const [code, state] = result.split("|");
+        if (state !== expectedState) throw new Error("STATE_MISMATCH");
+        return code;
+      } finally {
+        if (timeoutHandle) clearTimeout(timeoutHandle);
+      }
+    },
+    close: () => server.close(),
```
0xIchigo: This stops accepting new connections but doesn't tear down idle keep-alive sockets. We should either set `Connection: close` on the responses or call `server.closeAllConnections()` (Node 18.2+) in `close`, so `helius login` returns immediately

helius-cli/src/commands/login.ts
```diff
@@ -1,50 +1,123 @@
...
+        const label = identity?.kind === "wallet"
+          ? `wallet ${identity.display}`
+          : (identity?.display ?? "your account");
+        if (options.json) {
+          outputJson({ status: "ALREADY_LOGGED_IN", identity: identity ?? null });
+          return;
+        }
+        console.log(chalk.green(`✓ Already logged in as ${chalk.cyan(label)}.`));
+        return;
+      } catch {
```
0xIchigo: The already-logged-in probe falls through to a full browser re-login on any `listProjects` failure, including transient network errors or a 5xx. We should consider only falling through on auth failures and rethrowing / surfacing others

0xIchigo (changes requested):
Based 🔥 

Just have a few comments to resolve and then we're good to go

helius-cli/src/lib/loopback-server.ts
nathanliow: Fixed in 2f67c67. `state` is now validated inside the request handler — on a mismatch we serve the error page + 400 and reject with `STATE_MISMATCH`, so the browser reflects the real outcome and an early bad hit can't poison the flow. `startLoopback(expectedState)` takes the expected value up front.

helius-cli/src/lib/loopback-server.ts
nathanliow: Fixed in 2f67c67. Eliminated the join/split entirely — since `state` is now validated in the handler, `awaitCallback` resolves the raw `code` string directly. No more assumptions about the authorization code's charset.

helius-cli/src/lib/loopback-server.ts
nathanliow: Fixed in 2f67c67. `close()` now calls `server.close()` followed by `server.closeAllConnections()` (Node 18.2+), and every response sets `Connection: close`, so idle keep-alive sockets are torn down and `helius login` returns immediately.

helius-cli/src/commands/login.ts
nathanliow: Fixed in 2f67c67. The probe now only falls through to a fresh login on an auth failure: `if (classifyError(probeError).errorCode !== "INVALID_API_KEY") throw probeError;`. Transient failures (network/5xx/rate limit) are rethrown and surfaced. This required teaching `classifyError` to read the auth SDK's `API error (NNN)` message format (it previously fell through to `SDK_ERROR`), which also fixes exit-code classification for the other auth-SDK paths (signup/projects/oauth). Added two tests: auth-reject falls through to a fresh login; a 500 probe surfaces as exit 57 without re-logging in.

0xIchigo (dismissed):
LGTM 🔥

helius-cli/src/lib/loopback-server.ts
```diff
@@ -0,0 +1,106 @@
...
+  port: number;
+  awaitCallback: (timeoutMs: number) => Promise<string>;
+  close: () => void;
+}
+
+const SUCCESS_PAGE =
+  "<!doctype html><html><body style=\"font-family: -apple-system, system-ui, sans-serif; max-width: 480px; margin: 80px auto; text-align: center;\">" +
+  "<h1>Logged in</h1><p>You can close this tab and return to your terminal.</p></body></html>";
+
+const ERROR_PAGE = (msg: string) =>
```
0xIchigo, line 20: Please escape this before we merge—this is all unescaped HTML, and we pass the raw OAuth `error` query straight in, so there's a possibility of XSS

helius-cli/src/lib/loopback-server.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    // Serve an error page + reject, closing the connection so `helius login`
+    // returns immediately rather than waiting on a keep-alive socket.
+    const fail = (msg: string, err: Error) => {
+      res.writeHead(400, { "Content-Type": "text/html", Connection: "close" });
+      res.end(ERROR_PAGE(msg));
+      rejectCb(err);
+    };
+
+    if (error) {
```
0xIchigo, line 67: Tied to the comment above—this passes the attacker-controllable `error` value into the HTML page

helius-cli/src/commands/whoami.ts
```diff
@@ -0,0 +1,149 @@
...
+    const expired = expiresInSec !== null && expiresInSec <= 0;
+
+    let serverVerified: boolean | null = null;
+    if (options.verify) {
+      spinner?.start("Verifying token against backend...");
+      try {
+        await listProjects(jwt);
+        serverVerified = true;
+        spinner?.succeed("Token accepted by backend");
+      } catch {
```
0xIchigo: `--verify` treats every failure as `"Token rejected by backend"`, even if it's a transient error. `login.ts` solves this already as `INVALID_API_KEY` is the only real auth failure and then transient errors are re-thrown/surfaced

helius-cli/src/lib/browser.ts
```diff
@@ -0,0 +1,18 @@
+import { spawn } from "node:child_process";
+
+/**
+ * Opens `url` in the user's default browser. Detached + unref'd so the CLI
+ * doesn't wait on it. Returns false if the platform launcher couldn't spawn.
+ */
+export function openInBrowser(url: string): boolean {
+  const platform = process.platform;
+  const cmd = platform === "darwin" ? "open" : platform === "win32" ? "cmd" : "xdg-open";
+  const args = platform === "win32" ? ["/c", "start", "\"\"", url] : [url];
```
0xIchigo, line 22: This isn't exploitable today because the URL is fully app-constructed, but `cmd`'s handling of `&` and `start` is a footgun if a user-influenced URL ever reaches here. We should, at the very least, leave a comment about this invariant

helius-cli/CHANGELOG.md
```diff
@@ -0,0 +1,42 @@
+# Changelog
+
+> **⚠️ Before merging v2.0.0 / publishing to npm:** this release depends on
+> `oauthTokenExchange`, `HELIUS_API_URL` env override, and JSON error parsing
+> being available in `helius-sdk`. Those land in a separate SDK PR. Sequence:
+>
+> 1. Merge & publish the `helius-sdk` PR (new minor/major).
+> 2. Bump `helius-sdk` dep in `package.json` from `^2.2.2` to the new version.
```
0xIchigo: This is stale as we're bumping to 3.0.0

0xIchigo (changes requested):

helius-cli/src/lib/loopback-server.ts
nathanliow: Fixed in 4400a8d — added an `escapeHtml()` helper applied inside `ERROR_PAGE`, so every error message (including the OAuth `error` param) is escaped before rendering.

helius-cli/src/lib/loopback-server.ts
nathanliow: Fixed in 4400a8d — the escaping lives inside `ERROR_PAGE` itself, so the `fail(error, …)` path (and all other callers) renders the attacker-controllable `error` value escaped.

helius-cli/src/commands/whoami.ts
nathanliow: Fixed in 4400a8d — `--verify` now mirrors the `login.ts` probe pattern: only `INVALID_API_KEY` (401/403) marks the token rejected; transient errors (network, 5xx, rate limit) are re-thrown and surface via `handleCommandError` with their own classified exit code.

helius-cli/src/lib/browser.ts
nathanliow: Done in 4400a8d — documented the app-constructed-URL invariant (incl. the cmd /c start metacharacter footgun) and added an http(s)-protocol guard as defense-in-depth: non-http(s) or unparseable URLs return false before spawn.

helius-cli/CHANGELOG.md
nathanliow: Removed in 4400a8d — all three prerequisites are done (SDK 3.0.0 published, dep already at ^3.0.0, `tsc --noEmit` clean), so the whole pre-merge warning block was stale and is gone.

0xIchigo (approved):
LGTM :shipit:

PR #119, opened 2026-06-05
feat(ci): Require and Enforce Signed Commits Repo-wide

0xIchigo:
This PR makes signed commits an explicit, enforced requirement for all contributions, regardless of which package is touched (CLI, MCP, skills, plugin, cursor)


nathanliow (approved):
LGTM

PR #121, opened 2026-06-10
fix(cli): Store Config With Owner-only Permissions

0xIchigo:
This PR tightens permissions on `~/.helius/config.json` so the credentials it holds (i.e., a JWT session token and the Helius API key) are readable only by the owner. 

TLDR:
- `ensureDir()` creates `~/.helius` with mode `0o700`
- `save()` writes `config.json` with mode `0o600`
- `save()` also `chmod`s the file to `0o600` on every write, so configs created by installs predating this change get tightened on next save. Note that the `mode` write option only applies at file creation
- If chmod fails (e.g., non-POSIX filesystem), warn once to stderr with an actionable hint instead of silently leaving the file exposed

### DIFF
--- a/helius-cli/src/lib/config.ts
+++ b/helius-cli/src/lib/config.ts
@@ -79,7 +79,8 @@ interface Config {
 
 function ensureDir(): void {
   if (!fs.existsSync(CONFIG_DIR)) {
-    fs.mkdirSync(CONFIG_DIR, { recursive: true });
+    // Owner-only — the config holds a JWT and API key.
+    fs.mkdirSync(CONFIG_DIR, { recursive: true, mode: 0o700 });
   }
 }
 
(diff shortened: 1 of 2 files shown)


nathanliow:
Verified file behavior works (fresh creates `600`, loosened files re-tightened on save). One gap:

**The `mode: 0o700` in `ensureDir()` never applies in practice.** `feedback.ts` runs a modeless `mkdirSync(HELIUS_DIR)` at module load on fresh installs — before any command reaches `ensureDir()`, which then skips via `existsSync`. Tested with a fresh `HOME` through the real CLI: dir ends up `drwxr-xr-x` (config.json still `600`, so credentials stay protected — but the dir hardening in the PR description doesn't materialize).

Fix: add `mode: 0o700` to the `feedback.ts` mkdir, or chmod the dir in `ensureDir()` (also fixes existing installs).

Nit, out of scope: `keygen.ts` has the same modeless `mkdirSync`, and writes `keypair.json` before chmodding it.

nathanliow (approved):
LGTM

PR #126, opened 2026-06-15
fix(cli): exit non-zero when reclaim batches fail to land

mooncitydev:
hey, its moondev. was reading through helius-cli and hit a bug in `helius reclaim` that i think matters for anyone scripting against it.

### what's wrong

the batch send loop is continue-on-error (collects successes/failures, doesn't abort), which is the right call. but the summarize step at the end always reports success: in `--json` mode it calls `outputJson(...)` which emits `{ ok: true, ... }`, and in terminal mode it just prints and returns. either way the process exits 0.

so if every batch fails to broadcast (sender down, rate limited, tip too low, whatever), the command still exits 0 with `ok: true`. a script doing `helius reclaim --json && ...` treats a fully failed run as success, the rent never actually gets reclaimed, and nobody knows. same story for partial failures.

the readme pitches the cli as "built for developers and LLM agents to ... automate workflows", and agents/scripts key off the exit code and the `ok` flag, so this felt worth fixing.

### the fix

if any batch fails to land, exit non-zero (`PAYMENT_FAILED` / 22):

- `--json`: emit an error envelope (`ok: false`) with the full breakdown under `details` (closed count + per-batch successes and failures) so callers still see which batches landed
- terminal: keep the existing summary output, just exit non-zero at the end

fully successful runs are untouched (still `ok: true` / exit 0).

### tests

added `src/commands/reclaim.test.ts` covering all-fail, partial-fail, and all-succeed. ran the full helius-cli suite, 45 passing. commit is signed/verified per CONTRIBUTING.

not a huge change but should save someone a confusing debugging session. cheers

### DIFF
--- a/helius-cli/src/commands/reclaim.ts
+++ b/helius-cli/src/commands/reclaim.ts
@@ -9,6 +9,7 @@ import { formatSol } from "../lib/formatters.js";
 import {
   outputJson,
   exitWithError,
+  getExitCode,
   handleCommandError,
   createSpinner,
   withRetry,
@@ -400,21 +401,35 @@ export async function reclaimCommand(
       0,
     );
 
+    const summary = {
+      owner,
+      destination: String(destination),
+      region: options.region || "Default",
+      swqosOnly: !!options.swqosOnly,
+      batchSize,
+      totalBatches: batches.length,
+      closed: closedCount,
+      reclaimedLamports,
+      reclaimedSol: reclaimedLamports / 1_000_000_000,
+      successes,
+      failures,
+    };
+
     if (options.json) {
-      outputJson({
-        owner,
-        destination: String(destination),
-        region: options.region || "Default",
-        swqosOnly: !!options.swqosOnly,
-        batchSize,
-        totalBatches: batches.length,
-        closed: closedCount,
-        reclaimedLamports,
-        reclaimedSol: reclaimedLamports / 1_000_000_000,
-        successes,
-        failures,
-      });
-      return;
+      // A failed batch must never be reported as success: scripts and agents
+      // key off the exit code and the envelope `ok` flag. Emit a success
+      // envelope only when every batch landed; otherwise emit an error
+      // envelope (non-zero exit) that still carries the full breakdown under
+      // `details` so callers can see which batches are already on-chain.
+      if (failures.length === 0) {
+        outputJson(summary);
+        return;
+      }
+      exitWithError(
+        "PAYMENT_FAILED",
+        `${failures.length} of ${batches.length} reclaim batch(es) failed to land; ${closedCount} of ${toClose.length} account(s) closed.`,
+        summary,
+      );
     }
 
     console.log(chalk.bold("\nReclaim complete:\n"));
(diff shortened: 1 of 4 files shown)


helius-cli/src/commands/reclaim.ts
```diff
@@ -439,6 +454,9 @@ export async function reclaimCommand(
           "\n  Re-run the command to retry. Successful batches are already on-chain.",
         ),
       );
+      // Mirror the JSON path: a partial or total failure must exit non-zero so
+      // scripts wrapping the CLI don't treat it as a clean success.
+      process.exit(getExitCode("PAYMENT_FAILED"));
```
0xIchigo: The JSON path's `exitWithError` fires `sendCommandEvent(..., { success: false })`, but this raw `process.exit` doesn't, so the terminal-mode failures only record the preAction `success: true` event. We should set this to `sendCommandEvent(getCurrentCommand() ?? "reclaim", { success: false, exitCode })` before exiting to mirror the JSON path

0xIchigo: Nit: this also skips the postAction "Please run: helius feedback..." prompt on terminal failures, so it would be nice to have that in there as well

helius-cli/src/commands/reclaim.ts
```diff
@@ -400,21 +401,35 @@ export async function reclaimCommand(
...
+      // key off the exit code and the envelope `ok` flag. Emit a success
+      // envelope only when every batch landed; otherwise emit an error
+      // envelope (non-zero exit) that still carries the full breakdown under
+      // `details` so callers can see which batches are already on-chain.
+      if (failures.length === 0) {
+        outputJson(summary);
+        return;
+      }
+      exitWithError(
+        "PAYMENT_FAILED",
```
0xIchigo, line 436: `PAYMENT_FAILED` isn't in `RETRYABLE_CODES`, so this envelope reports `recoverable: false` while the message advises re-running. An agent strictly keying off `recoverable` won't auto-retry a transient batch (rate-limited / sender-down). We should resolve this

0xIchigo (changes requested):
Very based, thank you! I just have a few comments that need to be resolved before merging 🫡

mooncitydev:
okay 

1. **failure telemetry on terminal exits** - good catch. the raw `process.exit` halts before commander's postAction hook, so terminal-mode failures only logged the preAction `success: true` event. it now calls `sendCommandEvent(getCurrentCommand() ?? "reclaim", { success: false, exitCode })` before exiting, mirroring the json path via `exitWithError`.

2. **the skipped feedback prompt** - pulled that `Please run: helius feedback ...` string into a shared `printFeedbackPrompt()` in `output.ts` so the postAction hook and reclaim's terminal exit use one copy, and reclaim prints it before exiting too.

3. **recoverable:false vs "re-run"** - agreed, that was contradictory. added an optional `recoverable` override to `exitWithError` and set it `true` on the reclaim fail path, since re-running is safe (closed accounts get skipped on the next scan) and the batch failures are usually transient (rate-limited / sender down). `PAYMENT_FAILED` stays non-recoverable everywhere else for genuine payment failures.

added a terminal-mode test covering the `success: false` event and the prompt, plus `recoverable: true` asserts on the json fail paths. full suite green (46 tests).

0xIchigo (approved):
Thank you for your contribution! 🔥

PR #128, opened 2026-06-18
feat(ci): Harden Release Pipeline and Add Dependency/Security Automation

0xIchigo:
This PR fixes the CI/release gaps surfaced during the 2.0.0 releases and the codebase review via pure `.github/` config changes

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,30 @@
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: /helius-mcp
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 10
+    commit-message:
+      prefix: "chore(deps)"
+    groups:
+      mcp-minor-patch:
+        update-types: [minor, patch]
+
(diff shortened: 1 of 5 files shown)


nathanliow:
Signed-commit fix and version guards look correct. Two gaps before this fully hits the "Security Automation" goal:

**1. No npm provenance.** Both publish jobs run `pnpm publish` without `--provenance`/`id-token: write`, so tarballs aren't cryptographically tied to CI — the main thing that'd catch a malicious publish if `NPM_TOKEN` leaks.
```yaml
permissions:
  id-token: write
- run: pnpm publish --no-git-checks --access public --provenance
```

**2. `audit.yml` on `pull_request` blocks every PR** when any transitive prod dep gets a HIGH advisory, even docs-only changes. The Monday cron already covers us — suggest dropping the PR trigger (or making it non-blocking).

Nits: `mcp-release.yml` still on node 20 (rest moved to 22); "Open changelog PR" step missing `set -euo pipefail`.

nathanliow (approved):
LGTM

PR #133, opened 2026-06-18
feat(cli): Add Confirmation Prompts For Destructive Commands

0xIchigo:
This PR adds an interactive confirmation prompt before destructive CLI actions, since these commands were previously executed without confirmation. Each command also gains a `-y, --yes` flag to skip the prompt.

The following commands were gated:

| Command | Why it's destructive |
|---|---|
| `webhook delete <id>` | Deletes a webhook server-side, which is irreversible |
| `logout --all` | Wipes the entire local config (`jwt`, `apiKey`, `projectId`, `network`, `owsWallet`) |
| `stake unstake <acct>` | Signs a deactivate transaction with a user's keypair |
| `stake withdraw <acct>` | Signs a withdraw transaction with a user's keypair |

Note that it prompts only for an interactive human and auto-proceeds (i.e., no prompt) when any of these hold:

| Context | Prompt? |
|---|---|
| `--yes` | No |
| `--json` | No |
| Agent mode (`NO_DNA`) | No |
| Non-TTY (CI / pipes) | No |
| Interactive TTY, no flags | **Yes** (`y`/`yes` proceeds, anything else cancels) |

### DIFF
--- a/helius-cli/bin/helius.ts
+++ b/helius-cli/bin/helius.ts
@@ -255,6 +255,7 @@ program
   .command("logout")
   .description("Sign out and clear the local session token (keeps API key by default)")
   .option("--all", "Wipe the entire local config (jwt, apiKey, projectId, network, owsWallet)")
+  .option("-y, --yes", "Skip confirmation prompt (only relevant with --all)")
   .option("--json", "Output in JSON format")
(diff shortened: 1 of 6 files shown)


nathanliow (approved):
LGTM

PR #134, opened 2026-06-19
feat(cli): Add Confirmation Prompt Before Submitting Transactions

0xIchigo:
This PR is a follow-up to the destructive-command confirmation work added in #133 that adds a confirmation prompt before the `send` subcommands that actually submit a transaction on-chain, closing the gap noted there (the stake commands only sign; `send` is what hits the network)

The following commands are gated with `-y, --yes`:

| Command | Action |
|---|---|
| `send broadcast <tx>` | Broadcasts a signed transaction |
| `send raw <tx>` | Sends a raw transaction |
| `send sender <tx>` | Submits via Helius Sender |

`send poll` (i.e., read-only confirmation polling) and `send compute-units` (i.e., simulation only) are not gated

### DIFF
--- a/helius-cli/bin/helius.ts
+++ b/helius-cli/bin/helius.ts
@@ -1214,6 +1214,7 @@ sendCmd
   .command("broadcast <base64-tx>")
   .description("Broadcast a signed transaction and poll for confirmation")
   .option("--dry-run", "Preview without submitting the transaction")
+  .option("-y, --yes", "Skip confirmation prompt")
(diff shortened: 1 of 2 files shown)


nathanliow (approved):
Clean, tightly-scoped follow-up to #133. Reuses the `confirmDestructive` helper verbatim, correct prompt placement (after dry-run, before submission), and scoping matches the description (only broadcast/raw/sender gated; poll/compute-units left ungated). Build passes. LGTM.

PR #135, opened 2026-06-19
feat(mcp): Add `simulateTransaction` Tool For Pre-Send Transaction Dry-Runs

0xIchigo:
This PR adds a `simulateTransaction` tool to the Helius MCP. Given a base64-encoded transaction, it dry-runs it against the current chain state and returns success/failure, program error, compute units consumed, program logs, return data, and optional post-simulation account state, without submitting anything on-chain

This is the agent-facing counterpart to the human `--dry-run`/confirmation work shipped for the CLI: it enables the build -> simulate -> inspect -> send loop that autonomous agents need

Previously, the only simulation-adjacent capability was the CLI's `send compute-units`, which is a compute-unit estimate only; the MCP had nothing

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -125,6 +125,7 @@ Each skill has:
 - Always include `maxRetries: 0` when using Sender
 - Always include a Jito tip (minimum 0.0002 SOL) and priority fee
 - Use `heliusChain` + `getPriorityFeeEstimate` to get fee levels — never hardcode fees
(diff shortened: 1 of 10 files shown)


nathanliow (approved):
Reviewed: correct SDK call path (verified against helius-sdk's merged proxy → auto-send raw RPC), mutual-exclusivity of sigVerify/replaceRecentBlockhash enforced server-side, bigint-safe formatting, full router wiring, and complete error handling. 27/27 tests pass and tsc is clean. LGTM.

PR #136, opened 2026-06-22
feat(mcp): Add Timeout and Retry With Backoff For MCP RPC Reads

0xIchigo:
This PR makes the Helius MCP resilient to transient RPC failures for agent workloads. Today, the MCP has no retries and no timeouts, meaning that a 429 (rate limit) or 5xx failure on the tool call outright, and a slow/hung backend can block the tool indefinitely. This wraps the client so idempotent reads get a per-call timeout plus exponential-backoff retries, while writes, sends, and streaming are left untouched

This is introduced via a single `Proxy` wrapper applied once in `getHeliusClient()` with no per-tool changes. Resilience is applied only to methods matching `^(get|search|simulate)`:

| Call | Behavior |
|---|---|
| `get*` / `search*` / `simulate*` (DAS, raw RPC, `enhanced.getTransactions`, `webhooks.get/getAll`, `tx.getComputeUnits`) | 30s timeout + retry on 429/5xx/network |
| `sendTransactionWithSender`, `sendSmartTransaction`, `broadcastTransaction` | **untouched** — never retried, no external timeout |
| `webhooks.create/update/delete`, `pollTransactionConfirmation` | **untouched** |
| `ws` (streaming) | **untouched** |

### DIFF
--- a/helius-mcp/src/utils/helius.ts
+++ b/helius-mcp/src/utils/helius.ts
@@ -1,6 +1,7 @@
 import { createHelius, type HeliusClient } from 'helius-sdk';
 import { MCP_USER_AGENT } from '../http.js';
 import { getSharedApiKey } from './config.js';
+import { wrapClientWithResilience, withResilience, READ_TIMEOUT_MS } from './resilience.js';
 
 let sessionApiKey: string | null = null;
 let sessionNetwork: 'mainnet-beta' | 'devnet' = 'mainnet-beta';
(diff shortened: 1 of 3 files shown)


nathanliow:
Clean wrapper, correct send/streaming exclusions. One gap before merge:

**`wallet` and `zk` reads aren't wrapped.** `WRAPPED_NAMESPACES` omits them, so `client.wallet.*` (backs `heliusWallet`) and `client.zk.*` (backs `heliusCompression`) return unwrapped — no timeout, no retry. Both are pure-read namespaces, the exact case this PR targets. One-line fix:

```ts
const WRAPPED_NAMESPACES = new Set(['enhanced', 'webhooks', 'tx', 'wallet', 'zk']);
```

Add a test pair for each (mirroring `tx`) and update the description table.

Minor, non-blocking: SDK-method timeouts race instead of abort (socket leaks until the SDK's own timeout); only `restRequest` aborts.

nathanliow (approved):
Reviewed: clean single-Proxy design, correct read/write discrimination (verified against SDK 3.0.0 types — wallet/zk are pure-read namespaces, sends never match the retry regex), fail-fast on client timeouts, 251/251 tests green, typecheck clean. Minor hardening notes (restRequest GET-only doc, SDK-path timeout doesn't abort in-flight) are non-blocking follow-ups.

PR #137, opened 2026-06-24
feat(mcp): Refactor Regex-Parsing Output to Structured Continuation Metadata

0xIchigo:
The router decided "is there a next page?" by regex-parsing the handler's human-readable output — `**Next Page Token:** \`<token>\`` and `✅/❌ <signature>` lines (`detectContinuation` in `dispatch.ts`). The code itself flagged both patterns as "Fragile": any reformat of a handler's markdown silently breaks `expandResult` continuation, with no fallback and no test coverage

This PR replaces output-parsing with structured metadata, as handlers now emit a typed `ContinuationHint` on their response, and the router reads it directly

| File | Change |
|---|---|
| `results/types.ts` | New `ContinuationHint` type: `{ kind: 'paginationToken', api: 'history'\|'raw', token }` or `{ kind: 'signaturesQuick', lastSignature, until? }` |
| `utils/errors.ts` | `mcpText(text, { continuation })` + `continuation?` on the response type |
| `router/action-handlers.ts` | `continuation?` on `ActionHandlerResponse` |
| `tools/transactions.ts` | `getTransactionHistory` (parsed / raw / full-signatures / signatures-quick) and `getTransfersByAddress` emit the hint at each paginated return site |
| `router/dispatch.ts` | `detectContinuation` reads the hint (**both regexes deleted**); the hint is threaded `normalizeActionResponse → normalizeSuccessResponse → buildStoredResult`, and from `expandStoredResult`'s re-invocation so multi-page paging is preserved |

### DIFF
--- a/helius-mcp/src/results/types.ts
+++ b/helius-mcp/src/results/types.ts
@@ -7,6 +7,15 @@ export type TransactionHistoryContinuation =
   | { kind: 'historyApi'; paginationToken?: string }
   | { kind: 'rawApi'; paginationToken?: string };
 
+/**
+ * Structured continuation metadata emitted by an action handler on its response,
+ * describing how to fetch the next page. The router reads this directly instead
+ * of regex-parsing the handler's human-readable output.
+ */
+export type ContinuationHint =
+  | { kind: 'paginationToken'; api: 'history' | 'raw'; token: string }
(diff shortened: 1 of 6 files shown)


nathanliow (approved):
LGTM

PR #156, opened 2026-08-04
fix(mcp): Store Config With Owner-only Permissions

0xIchigo:
This bug was identified by Ray Sabee, [whitehatsecurity.nl](http://whitehatsecurity.nl/), or raysabee on GitHub

## Summary

`~/.helius/config.json` holds the user's JWT and Helius API key. `saveConfig` in `helius-mcp/src/utils/config.ts` wrote it with no mode and no `chmod`, so under the default umask (022) it was created `0644`, which is readable by every local user. `ensureDir` created `~/.helius` modeless for the same reason

`helius-cli` already fixed this in #121. `helius-mcp` never did, and since both packages share the same `config.json`, an MCP write would re-loosen a file the CLI had correctly locked down to `0600`

Reproduced against the pre-fix build under `umask 022`:

```
-rw-r--r--  1 user  staff  47  ~/.helius/config.json
```

After:

```
drwx------  3 user  staff  96  ~/.helius
-rw-------  1 user  staff  47  ~/.helius/config.json
```

## Changes

This PR mirrors the `helius-cli` implementation from #121 to store the config with owner-only permissions

- **`helius-mcp/src/utils/config.ts`** — `saveConfig` writes `{ mode: 0o600 }` plus an unconditional `chmodSync`, with a warn-once fallback if `chmod` throws (i.e., non-POSIX filesystems), so credentials are never left exposed silently. `ensureDir` creates at `0o700` and chmods unconditionally
- **`helius-mcp/src/utils/feedback.ts`** — `mkdirSync` at `0o700`. This runs at module load and could otherwise create `~/.helius` modeless before `config.ts` is reached. The CLI already had this.
- **`helius-mcp/src/utils/config.ts`, `helius-cli/src/commands/keygen.ts`** — keypair writes now pass `{ mode: 0o600 }`. Both were write-then-chmod, leaving a brief window where a freshly generated secret key was world-readable. The `chmod` is retained to cover overwrites of an existing file

`mode` only applies on create, which is why every site keeps both the `mode` and the `chmod`; the `chmod` is what tightens existing installs

## Testing

- New `helius-mcp/tests/config-permissions.test.ts` — 10 tests, and new `helius-cli/tests/config-permissions.test.ts` — 6 tests. Both run under a scratch `HOME` with `umask 022`, and refuse to run if the `HOME` override fails to take effect so they can never touch a real `~/.helius`. Cover fresh create, pre-existing `0644` file, pre-existing `0755` dir, load-path self-heal (`0644` and `0640`), already-`0600` no-op, and content round-trip
- Mutation-tested: reverting the source changes fails 4/10 (MCP) and 1/6 (CLI); reverting only the self-heal fails 2/10 and 1/6
- Order-independent: verified under `--sequence.shuffle`
- helius-mcp: 270/270 passing, `tsc` clean
- helius-cli: 63/63 passing, `tsc` clean
- Verified against the real built `dist` under `umask 022`, not just the harness

### DIFF
--- a/helius-cli/src/commands/keygen.ts
+++ b/helius-cli/src/commands/keygen.ts
@@ -43,7 +43,9 @@ export async function keygenCommand(options: KeygenOptions = {}): Promise<void> 
 
     // Save in Solana CLI format (64-byte array)
     const secretKeyArray = Array.from(keypair.secretKey);
-    fs.writeFileSync(resolvedPath, JSON.stringify(secretKeyArray));
+    // `mode` closes the window where a new keypair is briefly world-readable;
+    // the chmod still covers overwrites of an existing file.
+    fs.writeFileSync(resolvedPath, JSON.stringify(secretKeyArray), { mode: 0o600 });
     fs.chmodSync(resolvedPath, 0o600);
 
     // Get address for display
--- a/helius-cli/src/lib/config.ts
+++ b/helius-cli/src/lib/config.ts
@@ -116,9 +116,38 @@ function recoverConfig(raw: string): Config {
   return recovered;
 }
 
+/** chmod the config to owner-only, warning at most once per process if it fails. */
+function restrictConfigPerms(): void {
+  try {
+    fs.chmodSync(CONFIG_FILE, 0o600);
+  } catch {
+    // Non-fatal (e.g. non-POSIX filesystem), but the file may stay readable by
+    // other users — warn once so credentials aren't silently left exposed.
+    if (!warnedConfigPerms) {
+      warnedConfigPerms = true;
+      console.error(
+        `Warning: could not restrict permissions on ${CONFIG_FILE}; it may be ` +
+          `readable by other users. If possible, run: chmod 600 ${CONFIG_FILE}`,
+      );
+    }
+  }
+}
+
 export function load(): Config {
   if (!fs.existsSync(CONFIG_FILE)) {
     return {};
+  }
+
+  // Self-heal installs written before save() locked permissions down. Every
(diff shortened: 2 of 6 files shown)


NicolasPennie (dismissed):

nathanliow:
Verified locally: `helius-mcp` 266/266, `helius-cli` 57/57, `tsc` clean on both. Reverted `config.ts`/`feedback.ts` and re-ran — 4 of 6 new tests fail pre-fix. Grepped every `writeFileSync`/`mkdirSync`; all credential sites covered, `mode` + `chmod` pairing correct.

One gap worth a follow-up.

**Already-exposed installs never self-heal.** `loadConfig()` doesn't check permissions, and every `saveConfig` caller is an explicit user action (`auth.ts` signup/login/upgrade, `setSharedApiKey`, `savePreferences`) — nothing writes on startup, on a read, or on refresh. A user who set up once and only uses read tools keeps their `0644` JWT + API key indefinitely, which is the population the report describes. `loadConfig` already has a write-back path for corrupted configs:

```ts
try {
  if ((fs.statSync(SHARED_CONFIG_PATH).mode & 0o077) !== 0) {
    fs.chmodSync(SHARED_CONFIG_PATH, 0o600);
  }
} catch {}
```

`helius-cli`'s `load()` has the same gap — worth doing both together.

Nits: write-then-chmod leaves a negligible window when overwriting an existing `0644` file; the two "pre-existing" tests depend on the first test's file, so they'd `ENOENT` under `sequence.shuffle`; description says 5 tests, it's 6.

Not blocking — LGTM.

nathanliow (approved):
LGTM

PR #162, opened 2026-08-12
fix(mcp): Refuse Runtime Config Mutation Under a Shared Credential

0xIchigo:
## Why

`sessionApiKey`, `sessionNetwork`, and the memoized `heliusClient` are module-global. On stdio, that's right, as the process belongs to whoever launched it. Under one shared server-side credential serving many anonymous callers, any mutation is global

`setHeliusApiKey` would repoint the key for every concurrent caller, and `setSharedApiKey()` would write the caller-supplied value to `~/.helius/config.json`

This was already blocked in the common case. The handler returns early when `process.env.HELIUS_API_KEY` is set, which is how a hosted deployment supplies its key. The defect is that the guard is keyed on how the credential arrived rather than on the trust model; mount the key as a config file, let `getSharedApiKey()` find it, and the guard silently stops applying. This PR replaces incidental protection with intentional protection

## What

**Refuse mutation, at two layers.** `setApiKey()` and `setNetwork()` throw `SHARED_CREDENTIAL` via a shared `assertConfigurableAtRuntime()` helper. The `setHeliusApiKey` handler checks the mode first and returns a clean `UNSUPPORTED` error pointing the caller at `npx helius-mcp@latest` if they want their own key. The handler check gives a good message; the setter throw is the backstop for any future call site that forgets

**Fail fast at boot.** `HELIUS_MCP_SHARED_CREDENTIAL` set without `HELIUS_API_KEY` now exits non-zero. Previously the server started happily and answered every request with `NO_API_KEY`; a healthy-looking deployment serving errors, which is the shape of a bug you find during reviewer test cases instead of during deploy

**No signing key in shared mode.** Startup skips `loadKeypairFromDisk()` entirely. A server acting for many callers must not hold one, and the tools that would use it are off the hosted surface anyway

**No redundant seeding.** `getApiKey()` already resolves `HELIUS_API_KEY`, so shared mode doesn't copy env into session state, which is what made the old precedence order (`sessionApiKey` beats env) exploitable in the first place

## Testing

- `pnpm test` — 312 passing, 11 new in `tests/shared-credential.test.ts`. Covers: both setters refusing in shared mode and still working in single-tenant mode, the deployment key surviving a refused attempt, `HELIUS_API_KEY` and `HELIUS_NETWORK` resolving without runtime seeding, a missing key still reporting `NO_API_KEY` so boot can fail loudly, and the WebSocket URL returning a placeholder rather than the deployment key
- The WebSocket assertions cover code that landed in #160. They live here rather than there because they assert the shared-credential contract, which is what this file is about

### DIFF
--- a/helius-mcp/src/index.ts
+++ b/helius-mcp/src/index.ts
@@ -6,6 +6,7 @@ import { registerTools } from './tools/index.js';
 import { ROUTER_INSTRUCTIONS } from './router/instructions.js';
 import { setApiKey, setSessionSecretKey, setSessionWalletAddress } from './utils/helius.js';
 import { getSharedApiKey, loadKeypairFromDisk } from './utils/config.js';
+import { isSharedCredentialMode } from './utils/runtime.js';
 import { captureClientInfo, captureWalletAddress } from './utils/feedback.js';
 import { loadKeypair } from 'helius-sdk/auth/loadKeypair';
 import { getAddress } from 'helius-sdk/auth/getAddress';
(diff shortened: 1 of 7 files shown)


nathanliow:
The body says "no signing key in shared mode" — but only startup skips it. With
`HELIUS_MCP_SHARED_CREDENTIAL=1`, `loadSignerOrFail()` still loads the keypair
from disk (helius.ts:135), and `generateKeypair` still installs a process-global
signer (auth.ts:176). Confirmed both in isolation.

Also: `signup` calls `setApiKey()` at auth.ts:696, one line before the text
carrying the key — so shared mode returns a raw `SHARED_CREDENTIAL` error and
the caller never sees the provisioned key. Same at :711 and :723.

Fix: `isSharedCredentialMode()` guard at the top of `signup` and
`generateKeypair`, same as config.ts:27, plus a short-circuit in
`loadSignerOrFail()`. Or drop the claim from the body and link an issue.

Credential-mutation fix itself looks right.

nathanliow (approved):

PR #110, opened 2026-05-04
feat: compiler fan-out (plugin/cursor refs + --check) and CLI smoke tests

bernardogv:
## Summary

Two infra-hygiene improvements that share no files; review independently per commit.

- **Compiler fan-out** (`913db4d`): The compiler (`scripts/compile-skills.ts`) now keeps `helius-plugin/skills/<dir>/references/` and `helius-cursor/skills/<dir>/references/` byte-identical with `helius-skills/<skill>/references/`, and re-injects the `versions.json` version into plugin/cursor `SKILL.md` frontmatter. New `--check` mode diffs all generated outputs in-place and exits 1 on drift. Replaces the 6 duplicated bash blocks in `.github/workflows/sync-check.yml` (**592 → 22 lines**) with one `compile-skills.ts --check` step. Collapses 6 per-skill sync paragraphs in `CLAUDE.md` (**~210 → 103 lines**) to one short note.
- **CLI smoke tests** (`f1e235a`): Bootstraps `helius-cli` test infrastructure (vitest@2). Adds 16 smoke tests covering the JSON envelope contract from PRs #106/#107 — `classifyError` (Tier 2 status-in-message and Tier 3 keyword paths), `getExitCode` round-trip, `CLI_GUIDANCE` coverage, `jsonReplacer` bigint handling. Adds parallel `test-cli` job to `.github/workflows/test.yml`.

## Design intent

`SKILL.md` bodies in `helius-plugin/` and `helius-cursor/` remain **hand-managed** — they are intentionally not byte-identical to canonical (different MCP-prerequisite messaging, condensed router-surface section). The compiler now owns only the parts that should be identical (references) and the version field. This narrows the manual-sync surface while preserving the existing 'intentionally not identical' SKILL.md policy.

## Test plan

- [x] \`npx tsx scripts/compile-skills.ts\` — clean run, all 6 skills, no diff produced (no-op on current tree)
- [x] \`npx tsx scripts/compile-skills.ts --check\` — exit 0 on clean tree
- [x] Drift test (ref content): exit 1 with \`DRIFT: helius-plugin/skills/build/references/das.md\`
- [x] Drift test (cursor SKILL.md frontmatter): exit 1 with \`DRIFT: helius-cursor/skills/dflow/SKILL.md\`
- [x] Drift test (versions.json bump): exit 1 listing 10 affected files (canonical + .agents + mcp prompts + plugin & cursor SKILL.md)
[...]

PR #163, opened 2026-08-16
feat(scripts): Generate Plugin And Cursor Refs From The Compiler With `--check Mode`

0xIchigo:
This PR supersedes #110. Commit 1 is a cherry-pick of `913db4d` from @bernardogv's PR, so they remain its author; commit 2 addresses the review feedback

#110 could not merge as-is because both of its commits were unsigned, and `verify-signed-commits` requires GitHub-verified signatures on every commit, so the history needed rebuilding regardless. Its second changeset was also superseded: `helius-cli` has since picked up vitest@4, a `test` script, and a matrix entry in `test.yml`, so that commit's config, 855-line lockfile, and `test-cli` job now conflict for no gain. Only the compiler changeset is carried forward

## Why

`CLAUDE.md` documented six separate manual sync obligations, and `sync-check.yml` enforced them with 592 lines of bash containing six near-identical blocks. Adding a skill meant hand-writing another block; adding a destination meant editing all six

We're about to add a fifth destination for the ChatGPT/Codex plugin. Under the old setup, that's another ~100 lines of bash. Under this one, it's a table entry

## What

The compiler now owns everything generated from `helius-skills/`:

1. canonical `SKILL.md` frontmatter version, re-injected from `versions.json`
2. `.agents/skills/<skill>/` — Codex `SKILL.md` + 3 prompt variants
3. `helius-mcp/system-prompts/<skill>/` — the same variants, npm-shipped
4. `helius-plugin/skills/<dir>/` — refs bytewise, `SKILL.md` version re-injected
5. `helius-cursor/skills/<dir>/` — same

`--check` diffs every output in place and exits 1 on drift, replacing the bash from 592 to 22 lines. `CLAUDE.md` also gets reduced from 104 to 47

Plugin and cursor `SKILL.md` bodies stay hand-managed because they intentionally differ from canonical (different MCP-prerequisite messaging, condensed router surface). The compiler owns only what should be identical: the references and the version field. That preserves the existing "intentionally not identical" policy while shrinking the manual surface to just those bodies

## Review feedback addressed

- **Orphan detection.** `--check` walked canonical → dest only, so a reference renamed or removed upstream left its stale copy in every destination and every byte comparison still passed—weaker than the `comm -13` it replaced. New `reportOrphanRefs()` covers plugin, cursor, and `.agents`. We only report
- **Missing hand-authored `SKILL.md`** is now `MISSING:` drift under `--check`, a warning otherwise
- **`CLAUDE.md` workflow section removed.**
- **Not requested, same bug class:** a dest `SKILL.md` that exists but has no frontmatter was also skipped silently, so the version guarantee quietly stopped applying. Reported as `NO-FRONTMATTER:`

The two nits on `helius-cli` are moot—that changeset isn't carried forward. Its `tests/output.test.ts` is still worth porting separately; `classifyError`, `getExitCode`, and `jsonReplacer` in `src/lib/output.ts` have no coverage today

## Verification

| Case | Result |
|---|---|
| Clean tree, `--check` | exit 0, no drift or orphans |
| Clean tree, write mode | true no-op — zero files touched |
| Content drift in a plugin ref | `DRIFT: helius-plugin/skills/build/references/das.md`, exit 1 |
| Orphan in plugin refs | `ORPHAN: …/removed-upstream.md`, exit 1 |
| Orphan in `.agents` refs | `ORPHAN: .agents/skills/svm/references/…`, exit 1 |
| Missing dest `SKILL.md` | `MISSING:` + exit 1 under `--check`; warns and continues in write mode |
| `versions.json` bump | 10 files across all four destinations, incl. plugin + cursor `SKILL.md` |

Tree restored after each case. `helius-mcp` 320 tests and `helius-cli` 63 tests both pass 

Incidental: the old `CLAUDE.md` claimed DFlow had 12 reference files; the compiler reports 13. That stale count goes away with the per-skill paragraphs this replaces

Co-authored-by: Bernardo Gallegos-Vallejo <BernardoGVTech@gmail.com>


nathanliow (approved):
LGTM

PR #164, opened 2026-08-19
fix(plugin): pin helius-mcp to an exact version and drop the DFlow connector

nathanliow:
## Why

The plugin was removed from the **official** Claude marketplace during a curation pass. Nothing was flagged as malicious — Anthropic named exactly two blockers, both in `helius-plugin/.mcp.json`:

1. **Floating server version.** The marketplace pins this repo to a reviewed commit, but `"args": ["helius-mcp@latest"]` means the server code actually executed on install is whatever npm serves that day — outside that review.
2. **Unlisted third-party connector.** The plugin auto-connected to `https://pond.dflow.net/mcp`. An official listing requires third-party connectors to be in the Anthropic MCP connector directory, or dropped from the default config.

Both are closed here.

## What changed

**The two blockers**

- `helius-plugin/.mcp.json` and `helius-cursor/.mcp.json` now pin `helius-mcp@2.1.0` and declare only the `helius` server. Both files are 107 bytes and byte-identical.
- Both are generated by `scripts/compile-skills.ts` from a new `helius-mcp` key in `versions.json`. `--check` exits 1 on drift, so the config cannot silently regress to `@latest`.

**Why the pin lives in `versions.json` and not `helius-mcp/package.json`**

Deriving it from `package.json` would mean that between merging a version bump and publishing the release, `main` pins a version that does not exist on npm — breaking every install in that window. The pin is advanced deliberately *after* a publish lands. `sync-check` gained two guards:

- **blocking** (3 attempts, so a registry blip doesn't fail unrelated PRs): the pinned version must exist on npm.
- **warning only**: fires while the pin lags `package.json`. Warn rather than fail, because that lag is the correct state during the publish→pin window. Without it, a forgotten pin advance would sit green forever.

**Follow-on corrections** — dropping the connector makes these false, or they would mislead the listing review:

- DFlow prerequisite sections in both dflow skills rewritten to opt-in. The "isn't running, try restarting" recovery block is removed: absence is now the expected state, and that block would send users chasing a non-bug.
- "DFlow MCP Server — auto-starts with the plugin" corrected in both READMEs and the Cursor agent definition. That agent's stale "40+ live blockchain tools" is also corrected to the current 10-tool router surface, per `CLAUDE.md`'s Router Surface Maintenance rule.
- **Install commands were broken.** `/plugin marketplace add helius-labs/claude-plugins` points at a repo that does not exist, and `helius@helius` is not the marketplace name. Replaced with the community path (`helius@claude-community` — the manifest's `name` is `claude-community`, not `claude-plugins-community`) plus direct-from-source (`helius@helius-labs`).
- Plugin versions aligned at `1.2.0` across `marketplace.json`, `plugin.json`, and the Cursor manifest, resolving a pre-existing 1.0.0 / 1.1.0 / 1.0.0 three-way drift.
- `versions.json['helius-dflow']` 1.1.1 → 1.2.0. The shipped skill loses a prerequisite the agent acts on. Skill and plugin versions remain **independent namespaces** — both landing on 1.2.0 is coincidence, not a convention to preserve.

## Reviewing the diff

The `helius-dflow` bump makes the compiler restamp the version into every copy. These are **expected and compiler-owned — do not revert them, it will fail `--check`**:

- `helius-skills/helius-dflow/SKILL.md` — frontmatter version only, body untouched
- `.agents/skills/helius-dflow/` (SKILL.md + 3 prompts)
- `helius-mcp/system-prompts/helius-dflow/` (3 prompts)

No `references/*.md` file appears in the diff; those are byte-identical mirrors of canonical refs, which did not change.

## Verification

| Check | Result |
|---|---|
| `.mcp.json` size | 107 bytes, both files |
| `diff plugin/.mcp.json cursor/.mcp.json` | identical |
| Drift detection | reverting to `@latest` → exit 1, `DRIFT: helius-plugin/.mcp.json` |
| npm guard, pin `2.1.0` | exit 0 |
| npm guard, bogus `99.99.99` | exit 1 with `::error::` |
| Currency guard | silent (pin 2.1.0 == package.json 2.1.0) |
| `npx helius-mcp@2.1.0` | starts, reports `"version":"2.1.0"`, 10-tool surface |
| `helius-mcp` tests | 320 passed (12 files); catalog valid |
| Stale auto-start claims | 0 hits |
| Stale install commands | 0 hits |

Both CI guards were tested locally under `bash -e`, which is how GitHub Actions runs `run:` blocks.

## Merging this is not sufficient

The community marketplace entry pins:

```json
"url": "helius-labs/core-ai", "path": "helius-plugin",
"sha": "45cc2f0e8f8996f65a2a0c9a45cc1b63f1fde8c0"
```

That SHA is 36 commits behind `main` (2026-06-29), and `git show 45cc2f0:helius-plugin/.mcp.json` returns exactly the `@latest` + DFlow config fixed here. **Community installs keep shipping the old config until Anthropic advances that SHA** — the mirror repo is read-only, so it goes through `clau.de/plugin-directory-submission`. That submission, plus replying on the Anthropic thread with the merge commit, are the required follow-ups.

Incidentally, this entry is also what confirms `core-ai` is the reviewed repo.


0xIchigo (approved):
LGTM :shipit:

PR #165, opened 2026-08-19
fix(plugin): Finish the Router Migration and Single-source the MCP Pin

0xIchigo:
This PR is a follow-up to #164 that closes the gap that would have made the plugin look broken to the listing reviewer and made the pin impossible to half-apply

## Why

`jupiter`, `okx`, and `svm` were never migrated to the router surface. Their prerequisite gate reads:

> Check that Helius MCP tools are available (e.g., `getBalance`, `getAssetsByOwner`, `getPriorityFeeEstimate`).

Against the server #164 pins, none of those are tool names—they are actions under the 9 domain routers. Confirmed by handshake with `helius-mcp@2.1.0`:

```
tools: heliusAccount, heliusWallet, heliusAsset, heliusTransaction, heliusChain,
       heliusStreaming, heliusKnowledge, heliusWrite, heliusCompression, expandResult
```

So the check could never pass, and what follows is a hard stop:

> If they are NOT available, **STOP**. Do NOT attempt to call Helius APIs via curl or any other workaround.
> ```
> The Helius MCP server isn't running. Try restarting Claude Code.
> claude mcp add helius npx helius-mcp@latest
> ```

Install the plugin, run `/helius:jupiter`, and you are told the server is down and to install `helius-mcp@latest`—the floating tag #164 just removed;under the server name the plugin already registers. `build`, `dflow`, and `phantom` were migrated correctly; these three were missed; exactly the drift `CLAUDE.md`'s Router Surface Maintenance rule exists to catch

## What changed

**Gates and banners** — 12 files, `helius-plugin/` and `helius-cursor/`. Every gate names real tools (`heliusWallet`, `heliusAsset`, `heliusChain`; `heliusKnowledge` for svm). `jupiter`, `okx`, and `svm` gained the action-translation banner the other three already had

**The pin, single-sourced** — `scripts/compile-skills.ts` gained `pinMcpVersion()`, applied to canonical SKILL.md bodies, canonical `references/`, `install.sh`, the hand-managed plugin/cursor bodies, and all generated prompts. `stripClaudeSpecific` normalizes the token on the way in and re-stamps on the way out, so its patterns stay written against one literal. Scoped to the `helius-mcp` package; `helius-cli@latest` is a genuinely floating install and stays that way

### Why canonical is pinned, and not just the plugin

References are copied byte-for-byte from canonical into both packages, and CI enforces byte-identity. Pinning only the destinations would make them disagree with canonical by exactly that token and break the invariant that replaced ~420 lines of CI bash. So canonical is pinned, and standalone skills now also recommend the exact version. `install.sh` follows: it copies `SKILL.md` into the user's skills dir and then prints how to add the server, so a floating tag there would contradict the file it just installed

Deliberately excluded, and now documented in `CLAUDE.md`: `README.md`, `AGENTS.md`, `helius-skills/README.md`, `helius-skills/SYSTEM-PROMPTS.md`, `helius-mcp/README.md`, and the runtime recovery strings in `helius-mcp/src/`. Those describe adding `helius-mcp` standalone, outside any pinned plugin, where tracking latest is intended

**Versions** — all six skills bumped, since the shipped instruction text changed in each: `jupiter` and `okx` 1.0.1 → 1.1.0, `svm` 1.0.0 → 1.1.0 (behavior fix), `helius` 1.1.1 → 1.1.2, `helius-dflow` 1.2.0 → 1.2.1, `helius-phantom` 1.0.1 → 1.0.2 (pin token only). Plugin, Cursor manifest, and `marketplace.json` 1.2.0 → 1.2.1 — a bugfix release. New `helius-plugin/CHANGELOG.md`; the package being relisted had none.

## Reviewing the diff

Most of the 90 files are compiler-owned fan-out from the version bumps and the pin re-stamp. The hand-authored changes are:

- `scripts/compile-skills.ts`
- the 12 `SKILL.md` files under `helius-plugin/skills/` and `helius-cursor/skills/`
- `versions.json`, `CLAUDE.md`, both changelogs, three manifests

## Verification

| Check | Result |
|---|---|
| Hand-edit a pin → `--check` | exit 1, `DRIFT: helius-plugin/skills/jupiter/SKILL.md` |
| `.mcp.json` → `@latest` → `--check` | exit 1, `DRIFT: helius-plugin/.mcp.json` |
| Missing `helius-mcp` key | clean error, exit 1 |
| `helius-mcp@latest` in shipped skill content | 0 |
| `helius-cli@latest` | 5, untouched |
| Gate tool names vs live 2.1.0 surface | all 12 valid |
| `helius-mcp` build + tests | 320 passed (12 files) |
| npm pin guard / currency guard under `bash -e` | exit 0 / silent |


nathanliow (approved):
LGTM

PR #166, opened 2026-08-20
fix(plugin): Stop Instructing Third-Party Installs the Agent Shouldn't Run

0xIchigo:
## TLDR 

This PR is a follow-up to #164 and #165 to stop instructing third-party installs for agents 

## `curl | bash` in the okx skill

The prerequisite flow told the agent to run:

```
curl -fsSL https://raw.githubusercontent.com/okx/onchainos-skills/main/install.sh | bash
```

Piping a remote script into a shell is a standard curation flag, and the skill marks the OKX library **Required**, so this was normal setup rather than an edge case. `onchainos` is not published on npm under `onchainos` or `@okx/onchainos`, so the pipe is genuinely OKX's documented install, but the plugin does not have to embed it. It now points to OKX's instructions and tells the agent explicitly not to fetch and execute a remote install script on the user's behalf

**The same pattern shipped elsewhere and is fixed too:** `curl -fsS https://cli.dflow.net | sh` appeared in `references/dflow-agent-cli.md` and `references/integration-patterns.md` (both bytewise-mirrored into the plugin and Cursor), and `helius-skills/helius-okx/install.sh` echoed the OKX command as a next step, which would have contradicted the `SKILL.md` it had just installed

Tradeoff: the agent can no longer install these unattended, so those skills stall until the user acts. That is the correct direction for a curation review, and neither vendor's install path changed; it is just no longer embedded

## The DFlow connector was still paste-ready

`.mcp.json` no longer declares it, which satisfies the letter of the requirement. But the skill and README still handed over `claude mcp add --transport http DFlow https://pond.dflow.net/mcp`, so a reviewer grepping `pond.dflow.net` found the plugin actively instructing installation of the exact connector they raise

Both now state that it is a third-party hosted connector, that the agent must not configure it for the user, and point at DFlow's own setup docs

Also corrected a line that presumed the connector was present: "use the DFlow MCP server or DFlow docs" → "…if the user has added it, otherwise DFlow docs." Otherwise, the skill contradicted its own prerequisite section

## Compiler change

Deletes the DFlow rewrite rule in `stripClaudeSpecific`. It was anchored to the exact canonical sentence this PR removes, so it would have kept matching nothing, and `--check` cannot detect a rule that silently stops applying, since it only compares generated output against itself. I verified the non-Claude variants stayed free of `claude mcp add` and `restart Claude` afterward

## Reviewing the diff

37 files, +159/−243. Hand-authored:

- `helius-skills/helius-okx/SKILL.md`, `helius-plugin/skills/okx/SKILL.md`, `helius-cursor/skills/okx/SKILL.md`
- `helius-skills/helius-dflow/SKILL.md`, `helius-plugin/skills/dflow/SKILL.md`
- `helius-skills/helius-dflow/references/{dflow-agent-cli,integration-patterns}.md`
- `helius-skills/helius-okx/install.sh`, `helius-plugin/README.md`, `scripts/compile-skills.ts`
- `versions.json`, both changelogs, three manifests

The rest is compiler-owned fan-out — **do not revert it, `--check` will fail.**

Versions: `helius-dflow` 1.2.1 → 1.3.0 and `helius-okx` 1.1.0 → 1.2.0 (each loses an instruction the agent acts on); plugin, Cursor manifest, and `marketplace.json` 1.2.1 → 1.2.2

## Verification

| Check | Result |
|---|---|
| Remote-script pipes in shipped content | 0 |
| Paste-ready DFlow connector commands | 0 |
| `helius-mcp@latest` in shipped content | 0 |
| Reference byte-identity, 6 skills × 2 destinations | identical |
| `Anthropic MCP connector directory` in generated variants | 0 |
| Drift detection (`.mcp.json` → `@latest`) | exit 1, `DRIFT` |
| `install.sh` syntax, all 6 | pass |
| `helius-mcp` build + tests | 320 passed (12 files) |


nathanliow:
Verified at `74a33bf`. `--check` green, 320 tests pass. **`helius-plugin/` — the unit `marketplace.json` pins — is clean; nothing below blocks the listing.**

1. **`helius-plugin/skills/okx/SKILL.md:40`** — only reviewed-surface item. `npx skills add okx/onchainos-skills` is a live agent-executable third-party fetch (npm `skills` v1.5.23), four lines above `:47`'s "do not fetch and execute a remote install script."
2. **`helius-skills/helius-dflow/install.sh:74`** — still echoes `claude mcp add --transport http DFlow …`. Untouched; counterpart of the `helius-okx/install.sh` fix here.
3. **Cursor missed the DFlow change.** `grep -c "Do not configure it for the user"`: plugin 1, cursor 0. `skills/dflow/SKILL.md:34-40` keeps the paste-ready steps, `README.md:13` untouched, `CHANGELOG.md:6` claims otherwise.
4. **Three more dead `stripClaudeSpecific` rules** — same standard as the one you delete. Regexes in source order vs suspects-first: R3 `:220` 0→5, R6 `:242` 0→4 (R2/R1 eat their anchors), R4 `:227` 0→0 (unreachable, R3 subsumes it). Latent only. `:227` is the zero-risk delete.
5. **No guard.** `sync-check.yml`: 3 steps, 0 `grep` — #2 and #3 were uncatchable. One step failing on `curl .*\| *(bash|sh)` and `transport http DFlow` closes it.

Own issue: canonical `helius-okx/SKILL.md:35` + `helius-jupiter/SKILL.md:31` probe `getBalance`/`getAssetsByOwner`/`getPriorityFeeEstimate`; `router/register.ts` registers none of them → check unsatisfiable, agent always hits `STOP`.

nathanliow (approved):

PR #172, opened 2026-08-31
chore(mcp): Advance `helius-mcp` Pin to 2.2.0

0xIchigo:
This PR is the final step of the publish→pin sequence. `helius-mcp@2.2.0` is on npm, so the pin moves.

One hand-edited line:

```diff
-  "helius-mcp": "2.1.0",
+  "helius-mcp": "2.2.0",
```

The other 84 files are the compiler restamping the version token across both `.mcp.json` files, canonical and plugin/cursor `SKILL.md` bodies, `install.sh`, and the generated prompt variants—do not revert them, as `--check` will fail. The two remaining `helius-mcp@2.1.0` references are historical `1.2.0` changelog entries and correctly stay

## Verification

| Check | Result |
|---|---|
| `.mcp.json`, both packages | `helius-mcp@2.2.0`, byte-identical |
| Pinned version on npm | published |
| Currency guard | silent — pin matches `package.json` |
| Third-party install guard | pass |


nathanliow (approved):
LGTM

PR #173, opened 2026-08-31
fix(deps): Restore pnpm Overrides Dropped From Both Lockfiles

0xIchigo:
`main` has red Tests and Security Audit. Both have the same cause, and it's because we dropped the overrides from both lockfiles. This PR addresses that

## Cause

Merging #169, #170 and #171 deleted the `overrides:` block from both lockfiles. At `3174da7` `helius-mcp/pnpm-lock.yaml` had five entries; afterward, none. `helius-cli` lost its `ws` override too

`package.json` still declares them, so `pnpm install --frozen-lockfile` refuses to run:

```
ERR_PNPM_LOCKFILE_CONFIG_MISMATCH  The current "overrides" configuration
doesn't match the value found in the lockfile
```

Every workflow begins with that install, so Tests, Security Audit, and `MCP – Publish to npm` all fail before doing any work. The audit never reached `pnpm audit`

This is also a security regression: the overrides force patched `ws`, `hono`, `@hono/node-server`, `path-to-regexp` and `fast-uri`

## Fix

One authored line—the `fast-uri` override was stale independently, capping at `^3.1.2` while [GHSA-7p8r-x3mc-p8w7](https://github.com/advisories/GHSA-7p8r-x3mc-p8w7) widened to `<3.1.5`, permitting the vulnerable `3.1.4`:

```diff
-      "fast-uri@<3.1.2": "^3.1.2"
+      "fast-uri@<3.1.5": "^3.1.5"
```

Then `pnpm install --no-frozen-lockfile` in both packages to regenerate the lockfiles with overrides intact

## Verification

| Check | Before | After |
|---|---|---|
| `helius-mcp` frozen install | config mismatch | ✅ |
| `helius-cli` frozen install | config mismatch | ✅ |
| `helius-mcp` audit `--prod --audit-level high` | 1 high | ✅ 0 high |
| `helius-cli` audit | — | ✅ clean |
| `helius-mcp` tests | ❌ | ✅ 320 passed (vitest 4.1.11) |
| `helius-cli` tests | ❌ | ✅ 63 passed |
| `overrides:` in both lockfiles | absent | present |

The vitest 2→4 major bump in #171 is fine; 320 tests pass under 4.1.11


nathanliow (approved):
LGTM

PR #181, opened 2026-09-18
fix(helius-mcp): Label Mutating Actions as Writes

nathanliow:
`mutability` is defaulted from the public tool name, so four actions that mutate state are labelled reads. This PR fixes the labels and adds the predicate a hosted deployment will use to filter its surface. Note this PR only corrects catalogue metadata and adds an unused export — no registration, routing or runtime behaviour changes.

## The bug

`catalog.ts` defaults mutability by public tool name:

```ts
mutability: overrides.mutability ?? (publicTool === 'heliusWrite' ? 'write' : 'read'),
```

So these four are labelled `read`:

- `heliusStreaming.createWebhook`
- `heliusStreaming.updateWebhook`
- `heliusStreaming.deleteWebhook`
- `heliusAccount.purchaseCredits` — this one spends money

The label is silent until something filters on it, at which point the actions that most need protecting are the ones that slip through.

## What's here

- **Default is now `read`** rather than inferred, so a newly added action cannot quietly inherit a wrong value
- **The eight actions already in `mutationActions` declare the write explicitly.** That loop set `responseFamily: 'mutationReceipt'` but never touched `mutability` — the codebase already knew these mutate, it just said so in a different field
- **`purchaseCredits` gains `mutability: 'write'`**, matching its two immediate siblings `upgradePlan` and `payRenewal`, which both already carry it
- **`hostedEligible(entry)`** — filters on the auth requirement, not the mutability. A hosted multi-tenant server holds no wallet and no dashboard JWT, so `signer` and `jwt` actions are out; webhook CRUD stays in because it only ever needs the caller's own API key. `generateKeypair` and `setHeliusApiKey` are excluded by hand: one returns key material, the other mutates config for every caller
- **Three catalogue validations** so none of it drifts: every mutation receipt is a write, every `heliusWrite` action is a write, and the hosted surface is exactly 82 of 95
- **`tests/hosted-eligibility.test.ts`** pins the boundary — widening it silently is how a hosted server ends up signing on someone's behalf

## Explicitly not here

No registration filtering. Nothing calls `hostedEligible` yet; it is defined and tested so a later change can consume it. Zero behaviour change to the stdio build.

## Verification

- `pnpm build` clean
- `pnpm validate` passes, asserting the 82-action hosted surface
- 332 tests pass (320 existing, 12 new)

## Context

First PR of the hosted-MCP work. The immediate consumer is a no-replay policy on the Gatekeeper side: `mutability` is the field anyone would reach for to decide which actions are unsafe to retry, and it currently says these four are safe.


helius-mcp/src/router/catalog.ts
```diff
@@ -488,3 +492,41 @@ export function getActionsForTool(tool: RoutedPublicToolName): ActionName[] {
...
+
+/**
+ * Whether an action can be served by a hosted, multi-tenant deployment.
+ *
+ * The rule is the auth requirement, not the mutability: a hosted server holds no
+ * wallet and no dashboard JWT, so anything needing a `signer` or `jwt` is out.
+ * Webhook CRUD stays in — it mutates, but it only needs the caller's API key, so
+ * it works correctly under a bring-your-own-key deployment.
+ */
+export function hostedEligible(entry: ActionCatalogEntry): boolean {
```
0xIchigo: `authRequirement` is the wrong signal on its own. Several apiKey/none actions read or write host-local state and pass this predicate:

- `recommendStack` with `remember:true` calls `savePreferences` (`recommend.ts:231`), which writes `~/.helius/config.json`. That's the same file that got `setHeliusApiKey` excluded by hand, and it's also a mutation labelled read, which is exactly the no-replay case this PR is for
- `getAccountPlan` inherits none from the account-group default but calls `getJwt()` (`plans.ts:213`) and hits the dashboard, same path as the excluded `getAccountStatus`. `getHeliusPlanInfo` and `compareHeliusPlans` also read the JWT via `detectCurrentPlan` and print the operator's current plan
- `getStarted` checks the keypair file, reads the JWT, echoes `KEYPAIR_PATH` in its response (`auth.ts:142`), and every branch routes callers to excluded actions. No `isSharedCredentialMode` branch, unlike `generateKeypair` and signup

Suggest either adding these to `HOSTED_MANUAL_EXCLUSIONS` with a one-line reason each, or fixing the labels (`getAccountPlan` should be `jwt`; `recommendStack` should be write) so the predicate does the work

helius-mcp/src/router/catalog.ts
```diff
@@ -488,3 +492,41 @@ export function getActionsForTool(tool: RoutedPublicToolName): ActionName[] {
...
+  'generateKeypair',
+  'setHeliusApiKey',
+]);
+
+/**
+ * Whether an action can be served by a hosted, multi-tenant deployment.
+ *
+ * The rule is the auth requirement, not the mutability: a hosted server holds no
+ * wallet and no dashboard JWT, so anything needing a `signer` or `jwt` is out.
+ * Webhook CRUD stays in — it mutates, but it only needs the caller's API key, so
```
0xIchigo: `"Works correctly under a bring-your-own-key deployment"` describes a mode that doesn't exist yet. The only hosted mode is `HELIUS_MCP_SHARED_CREDENTIAL`, and `getApiKey()` (`helius.ts:59`) has no per-request key path. Under that model, any anonymous tenant can call `getAllWebhooks` to enumerate the operator's webhook URLs, then call `updateWebhook`/`deleteWebhook` on them. We could either scope the docblock and the test on line 51 to "valid only under a per-caller-key deployment, which does not exist yet", or exclude webhook CRUD until it does

helius-mcp/src/router/catalog.ts
```diff
@@ -488,3 +492,41 @@ export function getActionsForTool(tool: RoutedPublicToolName): ActionName[] {
...
+ */
+export function hostedEligible(entry: ActionCatalogEntry): boolean {
+  if (HOSTED_MANUAL_EXCLUSIONS.has(entry.action)) {
+    return false;
+  }
+  return entry.authRequirement === 'apiKey' || entry.authRequirement === 'none';
+}
+
+/** Size of the hosted surface. Exported so validation and tests agree on one number. */
+export const HOSTED_ACTION_COUNT = 82;
```
0xIchigo: A count fires on every ordinary read-action addition but stays silent on the case it's meant to catch: swapping one hosted action for another. Flipping a `jwt` action to none while adding `jwt` to any `apiKey` read keeps this at 82 and passes everything. A sorted inline snapshot of `getHostedActions()` names the entrant in the failing diff and subsumes the count

helius-mcp/src/router/catalog.ts
```diff
@@ -60,7 +60,7 @@ function makeEntry(
     aliases: overrides.aliases ?? [],
     authRequirement: overrides.authRequirement ?? 'apiKey',
     capabilityGate: overrides.capabilityGate ?? gate('agent', 'Available on every plan'),
-    mutability: overrides.mutability ?? (publicTool === 'heliusWrite' ? 'write' : 'read'),
+    mutability: overrides.mutability ?? 'read',
```
0xIchigo: `defaultDetail` and `handleEligibility` are already derived from `responseFamily` a few lines up. Deriving mutability the same way (`responseFamily === 'mutationReceipt' ? 'write' : 'read'`) leaves every label in this PR unchanged, makes the `mutability: 'write'` on line 184 redundant, and means a one-off `mutationReceipt` entry can't be a silent read until the validator runs. The three record-shaped writes already carry explicit overrides that `??` preserves

helius-mcp/src/scripts/validate-catalog.ts
```diff
@@ -73,6 +78,54 @@ for (const [key, product] of Object.entries(PRODUCT_CATALOG)) {
...
+  // Needing a signing key means acting as the wallet, which is a mutation
+  // wherever it appears. This is the check that would have caught `signup`,
+  // whose receipt shape and public tool both look like a read.
+  const needsSigner = entry.authRequirement === 'signer' || entry.authRequirement === 'jwtAndSigner';
+  if (needsSigner && entry.mutability !== 'write') {
+    error(entry.action, `requires "${entry.authRequirement}" but is labelled a read`);
+  }
+}
+
+// Assert the hosted surface from the catalogue rather than from the predicate
```
0xIchigo: This loop is a tautology. It rejects `signer`, `jwt`, `jwtAndSigner`, which is the exact complement of what `hostedEligible` admits on `catalog.ts:521`, so it can't fail for any catalog state without editing the predicate itself. The comment describes the independence the loop doesn't have. The independent checks (manual exclusions absent, `heliusWrite` absent) live only in vitest, so `pnpm validate` passes if `generateKeypair` is dropped from the exclusions alongside one new read. Either move those two checks here or delete this loop

helius-mcp/src/scripts/validate-catalog.ts
```diff
@@ -73,6 +78,54 @@ for (const [key, product] of Object.entries(PRODUCT_CATALOG)) {
...
+  }
+
+  if (entry.publicTool === 'heliusWrite' && entry.mutability !== 'write') {
+    error(entry.action, 'is a heliusWrite action but is labelled a read');
+  }
+
+  // Needing a signing key means acting as the wallet, which is a mutation
+  // wherever it appears. This is the check that would have caught `signup`,
+  // whose receipt shape and public tool both look like a read.
+  const needsSigner = entry.authRequirement === 'signer' || entry.authRequirement === 'jwtAndSigner';
```
0xIchigo: This `"which auth values imply a signer / local secret"` rule now appears five times as separate `===` chains (here, line 113, three test bodies, and the predicate's complement). A new `AuthRequirement` variant compiles cleanly and silently drops out of coverage. One exported classifier next to the type in `types.ts`, used everywhere, makes that a compile error

helius-mcp/tests/hosted-eligibility.test.ts
```diff
@@ -0,0 +1,109 @@
...
+
+import {
+  ACTION_CATALOG,
+  hostedEligible,
+  getHostedActions,
+  HOSTED_ACTION_COUNT,
+} from '../src/router/catalog.js';
+
+/**
+ * A hosted deployment is multi-tenant and holds no wallet, no dashboard JWT and
```
0xIchigo: `"Holds no dashboard JWT and no on-disk config"` is the assumption, but nothing enforces it. `getJwt()` is a bare `loadConfig().jwt` with no mode check, and no code in dispatch or registration references `hostedEligible`. The comment should say this is the intended contract, not the current runtime state, or a test should assert `"not hostedEligible implies shared-credential refusal at dispatch"`

helius-mcp/tests/hosted-eligibility.test.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  it('keeps webhook CRUD, which mutates but needs only the caller API key', () => {
+    for (const action of ['createWebhook', 'updateWebhook', 'deleteWebhook'] as const) {
+      expect(ACTION_CATALOG[action].mutability).toBe('write');
+      expect(hostedEligible(ACTION_CATALOG[action])).toBe(true);
+    }
+  });
+});
+
+describe('mutability labels', () => {
```
0xIchigo: Lines 60-66, 80-88, 25-39, and 98 re-assert validator checks that always run in the same `pnpm test` invocation. Keep the catalog-wide "for every entry" invariants in one layer, and keep this file for the boundary examples the validator can't express: the two manual exclusions, webhook CRUD, `getBalance` defaulting to read, and `heliusWrite` being excluded

helius-mcp/tests/hosted-eligibility.test.ts
```diff
@@ -0,0 +1,109 @@
...
+  });
+
+  it('defaults to read rather than inferring from the public tool name', () => {
+    expect(ACTION_CATALOG.getBalance.mutability).toBe('read');
+  });
+});
+
+describe('the hosted surface', () => {
+  it('matches the declared hosted action count', () => {
+    expect(Object.keys(ACTION_CATALOG)).toHaveLength(95);
```
0xIchigo: Duplicates the `ACTION_NAMES` completeness throw in `catalog.ts` and `tests/tools.test.ts:120`. We should use `ACTION_NAMES.length` here

0xIchigo (changes requested):

nathanliow:
All nine addressed in c519bda. Verified each claim against the code first — every one held.

**`authRequirement` was the wrong signal on its own.** Confirmed all four:

| Action | What it reaches |
|---|---|
| `recommendStack` | `savePreferences` → `saveConfig` → writes `~/.helius/config.json`, the same file that got `setHeliusApiKey` excluded |
| `getStarted` | reads the keypair file and JWT, echoes `KEYPAIR_PATH` at `auth.ts:142` |
| `getHeliusPlanInfo`, `compareHeliusPlans` | `detectCurrentPlan` → `getJwt()`, which is a bare `loadConfig().jwt` with no mode check |

All four are now excluded. `getAccountPlan` was the different case — mislabelled rather than missing an exclusion, since it calls `getJwt()` directly, so `jwt` is simply correct and the predicate removes it unaided. Exclusions carry their reason as data so the validator can quote it.

**Hosted surface: 82 → 78.**

**On the BYOK overclaim** — you're right, and it was the sharpest comment. Under the only hosted mode that exists, `getAllWebhooks` would enumerate the operator's webhooks for any anonymous caller. The docblock now states that the predicate is written for per-caller credentials, that they do not exist, and that nothing should call it until they do.

**Two adjustments to your suggestions, both from trying them:**

Deriving `mutability` from `responseFamily` does not make the `mutationActions` write redundant — the loop spreads the existing entry, which carries the already-derived `read` and overrides the derivation. It is restated for the same reason `defaultDetail` and `handleEligibility` are, now written down. Validation caught this immediately when I removed it.

`needsHostSecret` excludes `jwt` from the signer/write invariant. A JWT read is not a mutation, so folding all three into one check would have demanded `mutability: 'write'` on `getAccountStatus`.

**Second tautology — you were right again.** Rejecting signer/jwt/jwtAndSigner *is* the predicate's complement; I had replaced one tautology with another. The two independent checks (manual exclusions absent, `heliusWrite` absent) now live in the validator, so `pnpm validate` alone catches a dropped exclusion rather than relying on vitest.

**Snapshot over count**, demonstrated — swapping two actions prints:

```
+   "getAccountStatus",
+   "getStarted",
```

Also: `needsHostSecret` is one exhaustive `switch` in `types.ts`, so a new `AuthRequirement` variant is a compile error; tests trimmed to the boundary examples the validator cannot express; `ACTION_NAMES.length` replaces the repeated 95.

330 tests pass, `pnpm validate` clean.

